PR #47, opened 2015-02-14
Support TTC fonts

khaledhosny:
This is an initial attempt at TTC support (in preparation for TTC support in WOFF2). API changes are minimal; an optional `index` argument to `OTSContext::Process()` and otherwise no changes are needed to the clients. This is the simplest approach and has the obvious benefits, but it will probably means when a TTC font is used all the shared tables will duplicated which might increase memory usage.

CC @jfkthame, @jungshik, @behdad, @rsheeter

### DIFF
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -149,8 +149,9 @@ bool ProcessGeneric(ots::OpenTypeFile *header,
                     ots::Buffer& file);
 
 bool ProcessTTF(ots::OpenTypeFile *header,
-                ots::OTSStream *output, const uint8_t *data, size_t length) {
-  ots::Buffer file(data, length);
+                ots::OTSStream *output, const uint8_t *data, size_t length,
+                uint32_t offset = 0) {
+  ots::Buffer file(data + offset, length);
 
   // we disallow all files > 1GB in size for sanity.
   if (length > 1024 * 1024 * 1024) {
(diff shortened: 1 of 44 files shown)


rsheeter:
"all the shared tables will duplicated" - that sounds potentially unfortunate if the larger tables were to be shared. Sadly I don't have a good corpus of ttc files to give more specific data on how significant that is. Is avoiding that duplication thorny?

khaledhosny:
The other option is to give the client a complete TTC file and let them handle it, that will require some deeper changes to OTS, so I didn’t want to do that before checking with the browsers which API do they need/prefer.

behdad:
CC @jfkthame 

This is an interesting approach.  Jonathan, do you have any preference between this, and OTS sanitizing TTC as a whole?

src/ots.cc
```diff
@@ -233,6 +234,58 @@ bool ProcessTTF(ots::OpenTypeFile *header,
...
+
+  uint32_t num_fonts;
+  if (!file.ReadU32(&num_fonts)) {
+    return OTS_FAILURE_MSG_HDR("Error reading number of TTC fonts");
+  }
+  if (header->index >= num_fonts) {
+    return OTS_FAILURE_MSG_HDR("Requested font index is bigger than the number of fonts in the TTC file");
+  }
+
+  uint32_t offset_table_offsets[num_fonts];
```
anthrotype: variable length arrays is a C99 feature which is not supported in Microsoft Visual C++ 2010, so build fails on Windows with the following error:

    src\ots.cc(270): error C2057: expected constant expression
    src\ots.cc(270): error C2466: cannot allocate an array of constant size 0
    src\ots.cc(270): error C2133: 'offset_table_offsets' : unknown size

Perhaps you could replace line 270 with something like this:

    uint32_t *offset_table_offsets = new uint32_t [num_fonts];

Cheers,
Cosimo

khaledhosny: I reworked the code to remove the array entirely.

anthrotype: cool, thanks!

khaledhosny:
No response yet from Firefox or Chrome…

jfkthame:
Sorry for the slow response .... busy in non-font areas at the moment!

ISTM that the primary use case for TTC involves large shared tables -- particularly, sharing a large 'glyf' table between multiple variants of a CJK font. As such, I think it'd be preferable at least in some cases to decode the WOFF2 file back to a single, complete TTC, so that the client can benefit from this sharing if it's using more than one of the faces.

But there could also be cases where decoding a single face from the WOFF2-encoded TTC makes sense, so perhaps OTS should offer both types of API and let the client make the choice. E.g., if the WOFF2 file contains a TTC, and no index argument is provided, then decode the complete collection to a TTC resource; if an index is provided, decode a single face with no TTC header.

behdad:
SGTM.

khaledhosny:
Sounds reasonable, so I think I’ll go with that.

khaledhosny:
I have now implemented an API similar to what @jfkthame suggested. Not the prettiest code, but it seems to work.

khaledhosny:
I merged this with master now, if there are any issues we can deal with them later.

PR #51, opened 2015-03-12
Fix for C++11 compilation failure

jfkthame:
In my build on OS X (with clang), this fails with "error: invalid suffix on literal; C++11 requires a space between literal and identifier [-Wreserved-user-defined-literal]". (Adding the spaces improves readability, IMO, in addition to fixing the build failure.)


khaledhosny:
Thanks.

PR #53, opened 2015-03-16
[cff] Reject non-even args for hinting keys

samjhoward:
Non-even numbers of args result in silent failure in at least Chrome & Firefox on Win

### DIFF
--- a/src/cff.cc
+++ b/src/cff.cc
@@ -375,12 +375,18 @@ bool ParsePrivateDictData(
     operands.pop_back();
 
     switch (op) {
-      // array
+      // hints
       case 6:  // BlueValues
       case 7:  // OtherBlues
       case 8:  // FamilyBlues
       case 9:  // FamilyOtherBlues
-      case (12U << 8) + 12:  // StemSnapH (delta)
+        if (operands.empty() || ((operands.size() % 2) != 0)) {
+          return OTS_FAILURE();
+        }
+        break;
+
+	  // array
+	  case (12U << 8) + 12:  // StemSnapH (delta)
       case (12U << 8) + 13:  // StemSnapV (delta)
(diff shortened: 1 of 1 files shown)


khaledhosny:
Can you point to related Firefox /or Chrome bugs, test fonts and so on?

samjhoward:
Not sure it counts as a Firefox/Chrome bug since the spec does say that they should have an even number of args (Type 1 spec section 5.3) - since they're pairs of lower and upper bounds it wouldn't make sense to have an odd number anyway. 

I have a font which contains an odd number of args but I don't have the rights to send it over - would it help if I knocked together an example font with the issue?

Thanks,

Sam

khaledhosny:
I meant any references to any Firefox/Chrome bug reports concerning this issue, but a constructed test case should be fine as well.

khaledhosny:
Is this still an issue, do we have test files yet?

behdad:
I put a test file here:

http://behdad.org/NotoSansTC-Thin.otf.otf

It maps one character: U+3001, but with the hint corrupt to expose this issue.  I've put the non-corrupt file here:

http://behdad.org/NotoSansTC-Thin.otf.subset

samjhoward:
Apologies for the delay, I've been tied up with another project.

That's actually an example of a Charstring hinting command only having 1
arg, which is already tested for. The issue here is with the various blue
values keys in the Private DICT.

However, this has given me a handy (presumably) rights free font to modify
and generate a shareable test case.

http://jpedal.org/sam/hinting-issue/

The only difference between those should be one removed argument in the
first of the two private dictionaries. (And the offsets etc...)

Cheers,

Sam

khaledhosny:
Merged manually with some cosmetic changes, thaks.

PR #59, opened 2015-04-02
Make checksum calculation more resilient against stream seek

irori:
In cmap.cc, ots_cmap_serialise() may seek the output stream to a random
position in the table when it constructs UVS tables. But those seeks
didn't care about checksum calculation state of OTSStream, so it could
lead to an assertion failure in OTSStream::chksum().

After this patch, OTSStream::Write() simply accumulates checksum values
for incomplete words. As a result, it no longer needs to keep hanging
bytes.


khaledhosny:
Thanks.

PR #66, opened 2015-05-31
[collection] Take the offset into account for buffer length

gsnedders:
This fixes another crash (again found with afl) with collections. base64 `dHRjZgABAAAAAAAB/////w==` is enough to trigger the offset > length case, and `dHRjZgABAAAAAAABAAAABE9UVE8=` for the other.


khaledhosny:
Thanks for the fix, though the second one did not crash for me.

PR #67, opened 2015-06-02
Add Appveyor CI support

ebraminio:
There is an ongoing Appveyor CI support add on harfbuzz [here](https://github.com/behdad/harfbuzz/pull/112), don't know for sure but just guessed that can be useful for ots also and it is [working](https://ci.appveyor.com/project/ebraminio/ots/build/1.0.1), maybe we can upload the result somewhere so OTS binary would be easily accessible for average users without lot of hassle.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,17 @@
+platform:
+  - x64
+environment:
+  global:
+    MSYS2_BASEVER: 20150512
+  matrix:
+    - MSYS2_ARCH: x86_64
+install:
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks. Uploading the build should be fine as well, no idea how though.

anthrotype:
You can use Github Releases. You need to create a Github OAuth token add a new provider under the "deploy" section in your appveyor.yml setup file.

<http://www.appveyor.com/docs/deployment/github>

```yml
deploy:
  - provider: GitHub
    auth_token:
      secure: <here goes your encrypted github api key>
    draft: false
    prerelease: false
    on:
      appveyor_repo_tag: true       # deploy on tag push only
```

PR #76, opened 2015-08-31
Use google/woff2

khaledhosny:
This moves WOFF 2 support from our, more or less, duplicated copy of [google/woff2](https://github.com/google/woff2/) code to using the later as a submodule. It seems to work fine, and we get font collection support in WOFF 2 as a bonus.

Some fonts that is working in our code are broken now, but that is only because we have a partial fix here (see google/woff2#15).

Now that the WOFF 2 test suit is complete, this would allow us to handle any remaining issues in one place.


khaledhosny:
CC @behdad, @jfkthame, @rsheeter

khaledhosny:
The travis build failure is rather some ancient compilers, google/woff2#34 might fix it by hiding the offending line of code behind an `#ifdef`.

rsheeter:
Will look into merging google/woff2#34

PR #78, opened 2015-10-16
[deps] update brotli to origin/head (>0.2.0)

skomski:
ran into issues fuzzing ots related to https://code.google.com/p/chromium/issues/detail?id=538457


khaledhosny:
Thanks.

PR #81, opened 2015-10-30
Do not drop vhea/vmtx when GSUB does not exist

irori:
For context, see http://crbug.com/548124.

### DIFF
--- a/src/vhea.cc
+++ b/src/vhea.cc
@@ -4,7 +4,6 @@
 
 #include "vhea.h"
 
-#include "gsub.h"
 #include "head.h"
 #include "maxp.h"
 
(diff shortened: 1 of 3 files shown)


khaledhosny:
CC @jfkthame

khaledhosny:
Some tests will need to be updated as well, see the travis failures.

irori:
Tests updated.

khaledhosny:
@jfkthame is the original issue that required this check still relevant for Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=637481)?

behdad:
> @jfkthame is the original issue that required this check still relevant for Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=637481)?

Reading the issue, it sounds like it's a Windows limitation that we might want to address in OTS, for example by synthesizing a fake GSUB with vrt2 or something...

behdad:
> @jfkthame is the original issue that required this check still relevant for Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=637481)?
> 
> Reading the issue, it sounds like it's a Windows limitation that we might want to address in OTS, for example by synthesizing a fake GSUB with vrt2 or something...

Though, if the font already has GSUB, that will be complicated...

behdad:
Maybe this has been "fixed" in DWrite?

khaledhosny:
It sounds like a GDI-specific issue, but Firefox still support GDI.

jfkthame:
AFAICS from reading the issues, it seems reasonable to make this change. There's no fundamental reason for the presence of vhea/vmtx to be conditional on the presence of GSUB.

I don't think there's any need for OTS to try and do special magic here, like synthesizing a GSUB 'vrt2' feature if the font has vertical metrics but no GSUB. Such a font simply won't work very well on Windows, at least; not our problem.

khaledhosny:
Great. Merged manually.

PR #88, opened 2016-01-13
Includ a docs/HowToFix.md file

davelab6:
I'd like to document this information kindly gathered by Nicolas on the DFLT script font error that has come up recently.

### DIFF
--- a/docs/HowToFix.md
+++ b/docs/HowToFix.md
@@ -1,0 +1,61 @@
+# How To Fix
+
+This file describes the causes of common problems that OTS detects. 
+
+### DFLT script doesn't satisfy the spec
+
+    DFLT script doesn't satisfy the spec. DefaultLangSys is NULL
+    DFLT script doesn't satisfy the spec. LangSysCount is not zero: $lang_sys_count
+
+These errors are related to the language system declaration and how the lookup language tags are declared in the feature syntax.
+
+Old files made in FontLab Studio version 5.0, using Adobe FDK version 2.5 or so, then the `latin` language was the default, and thus could go unspecified. 
+And sometimes a language tag was declared after the untagged feature.
+This was a common practice to assume that script `latn` is the default script. 
+For example:
+
+```
+feature liga {
+   sub f f i by ffi;
+   sub f f l by ffl;
+   sub f f by ff;
+   sub f i by fi;
+   sub f l by fl;
+   language AZE;
+   language CRT;
+   language MOL;
+   language ROM;
+   language TRK;
+   script DFLT;
+   sub f f i by ffi;
+   sub f f l by ffl;
+   sub f f by ff;
+   sub f i by fi;
+   sub f l by fl;
+} liga;
+```
+
+But the script `latn` should be declared before language tags. 
+For example,
+
+```
+feature liga {
+   script DFLT;
+   sub f f i by ffi;
+   sub f f l by ffl;
+   sub f f by ff;
+   sub f i by fi;
+   sub f l by fl;
+   script latn;
+   sub f f i by ffi;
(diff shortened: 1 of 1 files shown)


behdad:
I'd rather we fix OTS to not reject these...

davelab6:
Since Firefox 44 will stop fonts with this issue from working, and its likely people will be searching the web for a fix, I thought this would be helpful :)

khaledhosny:
I don’t think the `DFLT` failure is bogus, but OTS has been “silently” dropping those tables from the start so people need to fix the fonts any way for old Chrome/Firefox releases. It is made more severe now since we are rejecting the whole font (at least for Firefox who doesn’t pass through the layout tables).

jfkthame:
Right, the DFLT failure is not a bogus check, the OpenType spec is quite clear on this.

There's one specific case where in principle, OTS could be adjusted to "fix" the font rather than rejecting it: if there's no DefaultLangSys, but there is a *single* explicit LangSysRecord under the DFLT script, then I think it would be reasonable for OTS to "migrate" that record to the DefaultLangSys, just issuing a warning.

And I suppose if there are multiple different LangSysRecords under DFLT *in addition to* a valid DefaultLangSys, then OTS could just drop those records, leaving only the default.

But in the case where there are multiple LangSysRecords and no DefaultLangSys, I don't see any basis for OTS to decide what should end up in the DefaultLangSys of a "fixed" font, and rejecting the font as ill-formed seems to be the only correct option. (And that's the case I see in real-world examples such as Merriweather Sans, unfortunately.)

behdad:
> Right, the DFLT failure is not a bogus check, the OpenType spec is quite clear on this.

Sure, it's quite clear.  But it's an arbitrary restriction.  It doesn't add any value, and neither Windows nor HarfBuzz rejects GSUB/GPOS because of this.  As such, I call the test wrong, not the other way around.

Indeed, for punctuation marks and other script=Common characters, I can easily imagine legitimate usecases where a LangSysRecord with loca feature is preferred in the DFLT script.

> There's one specific case where in principle, OTS could be adjusted to "fix" the font rather than rejecting it: if there's no DefaultLangSys, but there is a single explicit LangSysRecord under the DFLT script, then I think it would be reasonable for OTS to "migrate" that record to the DefaultLangSys, just issuing a warning.
> 
> And I suppose if there are multiple different LangSysRecords under DFLT in addition to a valid DefaultLangSys, then OTS could just drop those records, leaving only the default.
> 
> But in the case where there are multiple LangSysRecords and no DefaultLangSys, I don't see any basis for OTS to decide what should end up in the DefaultLangSys of a "fixed" font, and rejecting the font as ill-formed seems to be the only correct option. (And that's the case I see in real-world examples such as Merriweather Sans, unfortunately.)

behdad:
> There's one specific case where in principle, OTS could be adjusted to "fix" the font rather than rejecting it: if there's no DefaultLangSys, but there is a single explicit LangSysRecord under the DFLT script, then I think it would be reasonable for OTS to "migrate" that record to the DefaultLangSys,

That would be wrong.  For example, if that single explicit LangSysRecord is for Urdu, to convert digit 7 from Persian-style to Urdu-style, making it default will be wrong.

Really.  The restriction is bogus.  Both of you know extremely well how "authoritative" the spec should be considered...

>  just issuing a warning.

jfkthame:
> As such, I call the test wrong, not the other way around.

I think you mean "I call the *spec* wrong", not the test - yes?

In that case, the right thing to do is to propose a change to the spec, not to ignore what it says. (Who knows, maybe there was a good reason for it at the time? Maybe that reason is still relevant in some situation? Or maybe not....)

PR #89, opened 2016-01-14
INSTALL: Add how to run make to show src lines

davelab6:
This is generally useful so I suggest it go into the INSTALL


khaledhosny:
Merged as 841ac722d788c234fefcda556f296fa418e3346f.

PR #91, opened 2016-01-15
Fix #90

### LINKED ISSUE #90: Regression: missing overflow checks in src/cmap.cc
2cd7fabadbbfb559c3a7e9d8ca2dac0819f76f91 replaced an arbitrary upper limit check for cmap group size with comparing remaining buffer size. This isn't enough because it multiplies numbers from font without checking overflow. It causes huge memory allocation, resulting to crash. See https://code.google.com/p/chromium/issues/detail?id=577877 for more information.

bashi:
We should add overflow checks before multiplying group numbers in
src/cmap.cc to avoid huge memory allocation.


khaledhosny:
Thanks.

Issue #108, opened 2016-05-03
a font with sbix but without glyf should not be rejected


jungshik:
Currently, a font with both CBDT and CBLC (that are passed through) is allowed even when there's no glyf/loca table. 

That's not the case of a font with sbix table but without glyf table. One line fix would make this work, too.

behdad:
I'm actually not sure how good of an idea this is.  Apple's specification and possibly implementation expects sbix fonts to have glyf table.  FreeType implementation doesn't.

khaledhosny:
I didn’t know that, I just tested with FreeType. We should revert this then.

behdad:
Since Jungshik seems to have a font, maybe he can test with Apple.

jungshik:
In short, let's revert it.  Sorry that I hadn't tested on Mac before making a pull request (I meant to). I've just tested my hacked version of FruityGirl (only U+002[0A] supported) on Mac with Safari and Chrome (my build that passes through sbix).  

More details (still more testing is necessary): 

Safari does not like FruityGirl_sbix without glyf/loca when served as a webfont (no more details in Inspector/console).  Neither does Chrome on Mac (that passes a sbix font without glyf/loca; Blink gave me an decoding error after OTS passes it down to the OS). 

 OTOH, Safari does use the color bitmap for U+002A when FruityGirl_sbix with a nominal glyf table.  Moreover, its interpretation of BearingY seems to be different from the spec(FreeType's interpretation when tested inside Blink seems to match my reading of the spec.)  Chrome on Mac picks up a nominal outline in glyf table instead of using the bitmap in sbix table. 

I haven't yet tested either of them as a local font (as opposed to a web font).

jungshik:
FYI, attached is a zip file with two FruityGirl_sbix fonts  (hacked to cover only U+002[0A] in sbix). One has a nominal glyf table while the other does not. 

[FruityGirl_sbix.zip](https://github.com/khaledhosny/ots/files/247522/FruityGirl_sbix.zip)

khaledhosny:
OK, reverted.

behdad:
Thanks Jungshik.  Please file a bug against Chrome so we figure out what to do when glyf table is present.  Alternatively we might ask OTS to drop glyf table if sbix is present.  That's a gross hack :D.

How come Chrome works with Apple Color Emoji then?

jungshik:
[...]

PR #109, opened 2016-05-03
Do not reject a sbix font without loca/glyf tables.

jungshik:
This is for issue #108. 

See also http://crbug.com/605334 and https://codereview.chromium.org/1911123002


khaledhosny:
Thanks!

PR #130, opened 2017-07-21
Eliminate dynamic_cast so OTS can be built with -fno-rtti

### LINKED ISSUE #129: Eliminate the requirement for C++ RTTI
I notice that the new(ish) Table API now being used in OTS depends on RTTI being enabled, to allow the use of `dynamic_cast<...>`.

This is a problem at least for Gecko, which integrates OTS but is built with -fno-rtti. As such, the latest OTS code can't readily be dropped into Gecko (we're currently on 5.1.0, which predates the new Table API).

AFAICS, the use of RTTI is quite limited in scope, and I think it could be avoided without too much pain. Is this something you'd be willing to consider?

jfkthame:
Instead of relying on C++ RTTI, this adds an m_type field to Table, and each specific subclass constructor initializes this to the tag that corresponds to its type. This is used to implement a Font::GetTypedTable() method that checks whether the returned Table does in fact have the expected type (i.e. it wasn't loaded as unparsed data using TablePassthru), in which case it can safely be down-cast with static_cast<> by the caller.

Fixes #129.

### DIFF
--- a/src/cff.h
+++ b/src/cff.h
@@ -25,7 +25,7 @@ struct CFFIndex {
 class OpenTypeCFF : public Table {
  public:
   explicit OpenTypeCFF(Font *font, uint32_t tag)
-      : Table(font, tag),
+      : Table(font, tag, OTS_TAG_CFF),
         font_dict_length(0),
         local_subrs(NULL),
         m_data(NULL),
--- a/src/cff.cc
+++ b/src/cff.cc
@@ -972,8 +972,8 @@ bool OpenTypeCFF::Parse(const uint8_t *data, size_t length) {
     return OTS_FAILURE();
   }
 
-  OpenTypeMAXP *maxp = dynamic_cast<OpenTypeMAXP*>(
-      font->GetTable(OTS_TAG_MAXP));
+  OpenTypeMAXP *maxp = static_cast<OpenTypeMAXP*>(
+    GetFont()->GetTypedTable(OTS_TAG_MAXP));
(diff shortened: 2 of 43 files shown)


src/cff.h
```diff
@@ -25,7 +25,7 @@ struct CFFIndex {
 class OpenTypeCFF : public Table {
  public:
   explicit OpenTypeCFF(Font *font, uint32_t tag)
-      : Table(font, tag),
+      : Table(font, tag, OTS_TAG_CFF),
```
khaledhosny: Can’t we just use `tag` here in place of `OTS_TAG_CFF` since they should be the same? Ditto in other places where we don’t pass 0 for the type.

jfkthame: Yes, that would work equally well. I was thinking that in principle, at least, we could have a Table subclass that handles more than one tag, in which case it would have a single canonical "Type" but potentially various tag values. So the tag and the type are (in theory) distinct; tag could be whatever the caller chooses to pass, while type is constant for a given subclass, so it made sense to me that it should be hard-coded here.

But in practice I don't think that happens: with things like hmtx/vmtx, which share their format, they're handled as distinct (but trivial) subclasses of OpenTypeMetricsTable.

khaledhosny: OK, lets merge this and change it later.

behdad:
You could also make GetTypedTable be a template that takes the tag and returns the correct structure type...

PR #133, opened 2017-07-24
Add support for Graphite tables

kahsieh:
Support parsing and sanity checking for Graphite tables (Gloc v1; Glat v1, v2, v3; Feat v1, v2; Silf v1, v2, v3, v4, v5; Sile v1; Sill v1). The Graphite library may fail with bad input, historically resulting in many critical security bugs. See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1369672


khaledhosny:
This is crashing GCC 4.8 (at least on Travis) for some reason, any idea what would be causing this?

khaledhosny:
I moved master to use Trusty on Travis, can you rebase and see if it helps anyway, otherwise we can try a less ancient version of GCC (5 or even 6), I don’t think we really care about GCC 4.x anymore.

khaledhosny:
Thanks, that is quite some work!

I’d rather have the lz4 code as submodule under `third_party/` similar to brotli and woff2 instead of copying the code directly, to make it easier to update the code later. We might also want to make the lz4 dependency conditional since Chrome does not support Graphite and they might not be interested in an extra dependency for code they do not use (unless they already have lz4 for some other reason).

I might not be able to review the code for a while bit, but since this is mostly for Firefox I’m happy to merge it if @jfkthame is happy with the code.

src/ots.h
```diff
@@ -187,9 +187,12 @@ bool IsValidVersionTag(uint32_t tag);
 #define OTS_TAG_CFF  OTS_TAG('C','F','F',' ')
 #define OTS_TAG_CMAP OTS_TAG('c','m','a','p')
 #define OTS_TAG_CVT  OTS_TAG('c','v','t',' ')
+#define OTS_TAG_FEAT OTS_TAG('F','e','a','t')
```
khaledhosny: I think we will need to make the tag macros case sensitive if we ever supported AAT tables (which include a `feat` table all lowercase), but lets worry about this when it happens.

jfkthame:
On 25/07/2017 09:22, Khaled Hosny wrote:

> I’d rather have the lz4 code as submodule under |third_party/| similar 
> to brotli and woff2 instead of copying the code directly, to make it 
> easier to update the code later.

Yes, this sounds like the right thing to do.

> We might also want to make the lz4 
> dependency conditional since Chrome does not support Graphite and they 
> might not be interested in an extra dependency for code they do not use 
> (unless they already have lz4 for some other reason).

Perhaps all the Graphite tables (and then the lz4 dependency) should be 
controlled by a configuration option.

> I might not be able to review the code for a while bit, but since this 
> is mostly for Firefox I’m happy to merge it if @jfkthame 
> <https://github.com/jfkthame> is happy with the code.

Thanks. I'm really glad this is being done, and would like to see it 
merged soon, once any immediate issues are worked out.

I suggest we address the points above, while this is still in @kahsieh's 
fork; and I'm hoping to do some more review of what's actually being 
validated, though I suppose that can continue either before or after 
merging.

khaledhosny:
> Perhaps all the Graphite tables (and then the lz4 dependency) should be controlled by a configuration option.

Yes, a build time option should be fine.

kahsieh:
Regarding the build failure, my quick Google search would seem to suggest that the error

> g++-4.8: internal compiler error: Killed (program cc1plus)

is due to the server running out of memory. In fact, an earlier version of this pull request (exact same code, different comments) passed all the tests. I'll move the lz4 code, rebase, and push again today.

khaledhosny:
I fixed the CI issue by switching to GCC 5, and made lz4 a sumbodule. Now remains making the Graphite support optional at build time and I’ll merge this.

kahsieh:
I added a configuration option ./configure --enable-graphite that enables Graphite table sanitization and the corresponding lz4 dependency at build time. It is currently disabled by default; if we want Graphite support enabled by default, that can be done (and then the user would have the option to compile with --disable-graphite).

Since I removed the code that allows Graphite tables to pass through unchecked, an OTS build without Graphite support will now trigger the default behavior for unknown tables (i.e., drop). I think this is okay since any application that can make use of Graphite tables should compile OTS with Graphite support anyway.

src/ots.cc
```diff
@@ -131,12 +135,14 @@ const struct {
   { OTS_TAG_VMTX, false },
   { OTS_TAG_MATH, false },
   // Graphite tables
+  #ifdef OTS_GRAPHITE
```
khaledhosny: Please don’t indent pre-processor directives.

src/ots.h
```diff
@@ -209,13 +206,19 @@ bool IsValidVersionTag(uint32_t tag);
...
 #define OTS_TAG_PREP OTS_TAG('p','r','e','p')
-#define OTS_TAG_SILE OTS_TAG('S','i','l','e')
-#define OTS_TAG_SILF OTS_TAG('S','i','l','f')
-#define OTS_TAG_SILL OTS_TAG('S','i','l','l')
 #define OTS_TAG_VDMX OTS_TAG('V','D','M','X')
 #define OTS_TAG_VHEA OTS_TAG('v','h','e','a')
 #define OTS_TAG_VMTX OTS_TAG('v','m','t','x')
 #define OTS_TAG_VORG OTS_TAG('V','O','R','G')
+// Graphite tables
+#ifdef OTS_GRAPHITE
```
khaledhosny: This does not need to be behind the `#ifdef`, we better always have the macro around just in case it is needed.

tests/test_bad_fonts_nonrecursive.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
khaledhosny: Why is this script needed?

src/ots.cc
```diff
@@ -881,13 +884,19 @@ bool Font::ParseTable(const TableEntry& table_entry, const uint8_t* data,
...
       case OTS_TAG_PREP: table = new OpenTypePREP(this, tag); break;
-      case OTS_TAG_SILE: table = new OpenTypeSILE(this, tag); break;
-      case OTS_TAG_SILF: table = new OpenTypeSILF(this, tag); break;
-      case OTS_TAG_SILL: table = new OpenTypeSILL(this, tag); break;
       case OTS_TAG_VDMX: table = new OpenTypeVDMX(this, tag); break;
       case OTS_TAG_VORG: table = new OpenTypeVORG(this, tag); break;
       case OTS_TAG_VHEA: table = new OpenTypeVHEA(this, tag); break;
       case OTS_TAG_VMTX: table = new OpenTypeVMTX(this, tag); break;
+      // Graphite tables
+      #ifdef OTS_GRAPHITE
```
khaledhosny: Ditto here.

configure.ac
```diff
@@ -28,11 +28,21 @@ AX_CHECK_COMPILE_FLAG([-fsanitize=undefined], [UBSAN_CFLAGS=-fsanitize=undefined
...
-			      sanitization @<:@default=no@:>@]),
-	      AC_DEFINE(OTS_DEBUG, 1, [Enable extra debugging information]),
-	      :)
+              AS_HELP_STRING([--enable-debug=@<:@yes/no@:>@],
+              [Enable extra debugging information for font
+               sanitization @<:@default=no@:>@]),
+              if test $enableval = yes; then
+                AC_DEFINE(OTS_DEBUG, 1, [Enable extra debugging information])
+              fi,
+              :)
```
khaledhosny: I prefer the old indentation.

khaledhosny:
Travis and AppVeyor config files should be updated to build with Graphite enabled.

src/ots.cc
kahsieh: Will fix

src/ots.h
kahsieh: This prevents improper compilation of the Graphite table sanitization source files. Please let me know if you would prefer the tags to be available nonetheless.

tests/test_bad_fonts_nonrecursive.sh
kahsieh: This script is the same as test_bad_fonts.sh except that it does not recursively enter directories inside ./tests/fonts/bad, and therefore does not test graphite fonts, which we do not expect a non-Graphite build to catch.

khaledhosny: I don’t like the duplication of the scripts, but this can be handled differently later.

src/ots.h
khaledhosny: I think the tags should be available unconditionally, there is no harm in having them.

behdad: Also, public header should not rely on build config macros.

configure.ac
kahsieh: I'm not entirely sure what the old indentation was; it showed up improperly for me due to the mixed tabs and spaces, which is why I changed it. I'll indent the help string further since I can see that that changed, but please let me know if there's something else.

khaledhosny: I prefer not having unneeded changes when possible, like changing tabs to spaces, but not a big deal. Just indenting back the help string should be fine.

src/ots.h
khaledhosny: That is an internal header actually, the public one is `include/opentype-sanitiser.h`.

tests/test_bad_fonts_nonrecursive.sh
kahsieh: I don't like it either, but unfortunately the existing script already takes an argument, and autoconf doesn't allow us to pass arguments to test scripts anyway. I suppose the way to handle this without code duplication would be to modify the interface of test_bad_fonts.sh and then have test_bad_fonts_nonrecursive.sh call test_bad_fonts.sh.

configure.ac
kahsieh: Okay. In particular, the lines you have highlighted would change anyway because I wrapped the AC_DEFINE in an if statement.

tests/test_bad_fonts_nonrecursive.sh
khaledhosny: You can pass environment variables, see `AM_TESTS_ENVIRONMENT` in `Makefile.am`.

kahsieh:
Should Graphite support be enabled by default then, or only in tests?

tests/test_bad_fonts_nonrecursive.sh
kahsieh: I'll look into that.

jfkthame:
On 26/07/2017 17:05, Kevin Hsieh wrote:
> Should Graphite support be enabled by default then, or only in tests?

IMO, it should be enabled by default, so that a "standard" OTS build is 
as complete as possible, with an option to disable it for clients that 
know they don't care about that aspect.

khaledhosny:
Enabled by default is fine for me.

kahsieh:
Graphite support is now enabled by default and can be disabled using ./configure --disable-graphite. Also addressed the issues above.

khaledhosny:
`--disable-graphite` does nothing now, but I’m fixing this.

kahsieh:
The most recent commit appears to break my local build.

configure.ac
```diff
@@ -27,23 +27,21 @@ AC_SUBST(ASAN_CFLAGS)
...
+	AC_DEFINE(OTS_DEBUG, 1, [Enable extra debugging information])
+])
+
+AC_ARG_ENABLE([graphite],
+	AS_HELP_STRING([--enable-graphite=@<:@yes/no@:>@],
+                       [Sanitize Graphite tables @<:@default=no@:>@]))
+AS_IF([test "x$enable_graphite" != "xno"], [
+	AC_DEFINE(OTS_GRAPHITE, 1, [Sanitize Graphite tables])
+])
+AM_CONDITIONAL(OTS_GRAPHITE, [test "x$enable_graphite" != "xno"])
```
kahsieh: `enable_graphite` is no longer set in the default case here, resulting in a linker error in my local build when I configure with no options. The build system attempts to compile the Graphite table sanitization source files without doing the `AC_DEFINE`.

khaledhosny: There was a typo `= "xyes"` instead of `!= "xno"`, it should build fine now. `enable_graphite` is defined by `AC_ARG_ENABLE`, no need to manually define it.

kahsieh: Looks good to me

khaledhosny:
I’m getting two [oss-fuzz](https://github.com/google/oss-fuzz) issues related to the new Graphite support but I don’t have time to look into them, anyone wants to be added to the OTS contacts to look into them?

kahsieh:
I can look into the issues.

khaledhosny:
I don’t know what is the exact process, but I guess you need to send a PR like [this](https://github.com/google/oss-fuzz/pull/696) adding you to OTS’s project.yaml with an [email linked to a Google account](https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-an-e-mail-associated-with-a-google-account) so that you can access the bug reports.

kahsieh:
PR submitted!

PR #137, opened 2017-08-03
[graphite] Additional fixes

kahsieh:
Address issues detected when integrating with Firefox


khaledhosny:
Why was `1254925.ttf` removed? If it passes now, then move it to the good fonts instead.

kahsieh:
With this patch, OTS repairs `1254925.ttf` instead of rejecting it. Should it still be placed with the good fonts (since it technically passes) even though it's a bad font?

khaledhosny:
Yes, good fonts are the ones we don't reject, whether repaired or not.

kahsieh:
This commit restores 3 bad (but sanitizable) fonts to `tests/fonts/good/graphite_sanitizable` as test cases, and moves 6 other bad cases which are no longer completely rejected into that same folder.

khaledhosny:
Do we really need to drop all Graphite tables if one is dropped? If so, I’d rather make it that Graphite tables depend on each other and dropping one will cause the others to be dropped, somewhat similar to how GPOS/GSUB might depend on GDEF.

src/ots.cc
```diff
@@ -938,6 +938,21 @@ Table* Font::GetTypedTable(uint32_t tag) const {
...
+void Font::DropGraphite() {
+  file->context->Message(0, "Dropping all Graphite tables");
+  for (const std::pair<uint32_t, Table*> entry : m_tables) {
+    if (entry.first == OTS_TAG_FEAT ||
+        entry.first == OTS_TAG_GLAT ||
+        entry.first == OTS_TAG_GLOC ||
+        entry.first == OTS_TAG_SILE ||
+        entry.first == OTS_TAG_SILF ||
+        entry.first == OTS_TAG_SILL) {
+      entry.second->Drop("Discarding Graphite table");
```
khaledhosny: It would be nice to also show the table tag in the message.

kahsieh: The table tag will be printed because `Drop()` is called off of the specific Table instance.

kahsieh:
There are interdependencies among the Graphite tables, and I hesitate to predict how the Graphite library will behave under all circumstances where it receives only a subset of Graphite tables. Perhaps @jfkthame understands more about this. In this patch, dropping a Graphite table will result in a call to `Font::DropGraphite()`, which will drop all Graphite tables.

jfkthame:
I think dropping all graphite tables if any of them is rejected makes sense in general. In principle, if a single graphite table were dropped, the engine should recognize the brokenness of the font and safely reject it, but there's little point in letting things get that far. (After all, in principle the engine should behave safely in the face of arbitrary malicious data, too, but we're still trying to add an extra layer of protection in case of any weaknesses in its implementation.)

One exception: if the 'Sill' table is invalid and dropped, I don't think it is necessary to drop the rest of the graphite tables. AFAIK, nothing else depends on this table; if it is absent, there just won't be any mapping of language tags to features.

kahsieh:
The Sill table note is a a good point, since Sill tables didn't even exist in earlier versions of Graphite. I will revise the patch accordingly.

khaledhosny:
Thanks!

PR #143, opened 2017-09-03
[graphite] Fix binary-search header fields rather than dropping table

jfkthame:
The correct values for these fields can be computed from the number of entries
in the list to be searched, so we can simply overwrite them with the proper
values rather than rejecting tables where they appear wrong.

(Existing fonts in circulation often have bad values in these fields; in
practice, it doesn't matter because the current engine does not rely on
them, and therefore the errors have not been noticed. But in case any future
engine does try to use them, we should fix them up here.)

### DIFF
--- a/src/silf.cc
+++ b/src/silf.cc
@@ -278,6 +278,9 @@ bool OpenTypeSILF::SILSub::ParsePart(Buffer& table) {
   if (!table.ReadU16(&this->numPseudo)) {
     return parent->Error("SILSub: Failed to read numPseudo");
   }
+
+  // The following three fields are deprecated and ignored. We fix them up here
+  // just for internal consistency, but the Graphite engine doesn't care.
   if (!table.ReadU16(&this->searchPseudo) ||
       !table.ReadU16(&this->pseudoSelector) ||
       !table.ReadU16(&this->pseudoShift)) {
@@ -285,7 +288,6 @@ bool OpenTypeSILF::SILSub::ParsePart(Buffer& table) {
   }
   if (this->numPseudo == 0) {
     if (this->searchPseudo != 0 || this->pseudoSelector != 0 || this->pseudoShift != 0) {
-      parent->Warning("SILSub: Correcting binary-search header for zero-length pseudos list");
       this->searchPseudo = this->pseudoSelector = this->pseudoShift = 0;
     }
   } else {
@@ -293,7 +295,6 @@ bool OpenTypeSILF::SILSub::ParsePart(Buffer& table) {
     if (this->searchPseudo != 6 * (unsigned)std::pow(2, floorLog2) ||
         this->pseudoSelector != floorLog2 ||
         this->pseudoShift != 6 * this->numPseudo - this->searchPseudo) {
-      parent->Warning("SILSub: Correcting binary-search header for pseudos list");
       this->searchPseudo = 6 * (unsigned)std::pow(2, floorLog2);
       this->pseudoSelector = floorLog2;
       this->pseudoShift = 6 * this->numPseudo - this->searchPseudo;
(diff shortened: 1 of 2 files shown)


jfkthame:
@kahsieh FYI, this is prompted by looking through issues we're hitting in the Awami font. I have asked Martin to confirm about the multiplication by the entry size, as currently the docs are inconsistent; we may need to follow up with a further fix here once this is clarified.

kahsieh:
I believe that for those values, the spec is inconsistent but existing fonts in circulation are consistent with what I wrote. No good fonts failed those checks in my testing, and failing those checks were indicative of some kind of misalignment in the font.

jfkthame:
The Awami font used in the last demo fragment at http://scripts.sil.org/cms/scripts/page.php?site_id=projects&item_id=graphite_fontdemo#8f082ec6 has bin-search header fields in the SIL_Pass structure that have the values multiplied by the entry size, e.g.:

```
  numRanges = 161
  searchRange = 768
  entrySelector = 7
  rangeShift = 198
```

This contrasts with the older fonts (such as Padauk) used earlier on the same page, which have unmultiplied values, e.g.:
```
  numRanges = 185
  searchRange = 128
  entrySelector = 7
  rangeShift = 57
```

Because Awami is the newest of the fonts used there, I took this to indicate that the compiler has (relatively recently) been fixed to apply the multiplication, and the older fonts are generated by an old compiler that inadvertently failed to do this. As the engine does not actually use these fields, the error went unnoticed.

(I've asked for clarification of this from the Graphite developers, so when I hear back we'll be able to confirm that we are fixing this in the "right" direction. But given the discrepancy between old and new fonts, both of which are expected to work, we need to do something other than just reject here.)

kahsieh:
I see. It's also possible that we should do the check and/or fix differently based on the Silf table version here, but it'll depend on what the Graphite developers say. With regard to the font on that page, I'm still trying to figure out why the lz4 we're using won't decompress it (will probably have to compare to Graphite).

jfkthame:
A simple fix to make it decompress successfully is to replace the use of `LZ4_decompress_safe` with `LZ4_decompress_safe_partial` (and pass `decompressed.size()` again as the extra final arg). But I don't fully understand why plain `LZ4_decompress_safe` is failing, and I'd like to get to the bottom of that before just working around it.

src/silf.cc
```diff
@@ -278,19 +278,26 @@ bool OpenTypeSILF::SILSub::ParsePart(Buffer& table) {
...
+    }
+  } else {
+    unsigned floorLog2 = std::floor(std::log2(this->numPseudo));
+    if (this->searchPseudo != 6 * (unsigned)std::pow(2, floorLog2) ||
+        this->pseudoSelector != floorLog2 ||
+        this->pseudoShift != 6 * this->numPseudo - this->searchPseudo) {
+      parent->Warning("SILSub: Correcting binary-search header for pseudos list");
+      this->searchPseudo = 6 * (unsigned)std::pow(2, floorLog2);
+      this->pseudoSelector = floorLog2;
+      this->pseudoShift = 6 * this->numPseudo - this->searchPseudo;
```
kahsieh: The spec doesn't seem to have multiplication here?

kahsieh: On SILSub::pseudoShift, that is.

jfkthame: Right, but that's clearly a spec bug -- it would make the computation of pseudoShift meaningless. I've raised this with the developers, and assume the spec will get fixed.

jfkthame:
The Graphite table spec has just been updated to deprecate these fields, given the existing inconsistencies; the engine will never make use of them. Accordingly, I have removed the warning messages here. The code will now silently fix the values (just for internal consistency), but as far as the engine is concerned, it is irrelevant.

PR #145, opened 2017-09-04
[graphite] Allow for padding when decompressing

kahsieh:
The Awami font at http://scripts.sil.org/cms/scripts/page.php?site_id=projects&item_id=graphite_fontdemo has padding at the end of its Glat and Silf tables, so `table.remaining()` is not reflective of the size of the compressed data. By using `LZ4_decompress_safe_partial`, LZ4 will stop decompressing once it has produced the appropriate number of bytes and not complain about unconsumed input.


kahsieh:
cc @jfkthame

src/glat.cc
```diff
@@ -201,10 +201,11 @@ bool OpenTypeGLAT_v3::Parse(const uint8_t* data, size_t length,
...
       }
       std::vector<uint8_t> decompressed(this->compHead & FULL_SIZE, 0);
-      int ret = LZ4_decompress_safe(
+      int ret = LZ4_decompress_safe_partial(
           reinterpret_cast<const char*>(data + table.offset()),
           reinterpret_cast<char*>(decompressed.data()),
           table.remaining(),
+          decompressed.size(),
           decompressed.size());
       if (ret < 0) {
```
jfkthame: I think rather than just checking for `ret < 0` here, we should explicitly check that `ret == decompressed.size()` to confirm that decompression generated the expected amount of data; if not, then even though LZ4 didn't report an error, something is still badly wrong.

src/silf.cc
```diff
@@ -39,10 +39,11 @@ bool OpenTypeSILF::Parse(const uint8_t* data, size_t length,
...
         }
         std::vector<uint8_t> decompressed(this->compHead & FULL_SIZE, 0);
-        int ret = LZ4_decompress_safe(
+        int ret = LZ4_decompress_safe_partial(
             reinterpret_cast<const char*>(data + table.offset()),
             reinterpret_cast<char*>(decompressed.data()),
             table.remaining(),
+            decompressed.size(),
             decompressed.size());
         if (ret < 0) {
```
jfkthame: As above.

jfkthame:
BTW, did you determine exactly what padding is present -- is it just 0-3 bytes to pad the end of the table to a 4-byte boundary, or is there something more?

kahsieh:
Yes, that's exactly it. The Glat table has 2 bytes of padding and the Silf table has 3 bytes of padding.

kahsieh:
I changed the drop condition to `if (ret != decompressed.size())`.

khaledhosny:
@jfkthame, @kahsieh: please let me know when this and the other pull requests are ready to merge.

kahsieh:
I believe they are ready to merge at this point.

khaledhosny:
Thanks!

PR #155, opened 2017-10-24
Avoid #define-ing htons and friends

irori:
`opentype-sanitiser.h` #defines `ntohl`, `ntohs`, `htonl` and `htons` on Windows.
This causes problems if other headers #included later (like Winsock2.h) tried to define these as functions.

To avoid polluting the global namespace, this patch adds `ots_` prefix to these macros.


irori:
Context: https://crbug.com/776946

bashi:
Not a project owner anymore, but +1 for this change.

PR #159, opened 2018-02-15
Fix typo in HowToTest.md (s/ckeck/check/)

dholbert:
I verified locally that the currently-recommended command, `make ckeck`, is unrecognized, whereas the corrected `make check` command *does* actually run the tests. :)


khaledhosny:
Thanks!

PR #161, opened 2018-02-16
Don't bail if CFF2 glyph table is present

### LINKED ISSUE #160: Allow passing through of CFF2 table
Compare 
https://chromium-review.googlesource.com/c/chromium/src/+/706248/11/third_party/ots/src/ots.cc

I suggest to add CFF2 to the passthrough glyph tables:

```
--- a/third_party/ots/src/ots.cc
+++ b/third_party/ots/src/ots.cc
@@ -702,9 +702,11 @@ bool ProcessGeneric(ots::FontFile *header,
       // We don't sanitize bitmap tables, but don’t reject bitmap-only fonts if
       // we are asked to pass them thru.
       // Also don’t reject if we are asked to pass glyf/loca thru.
-      if (!font->GetTable(OTS_TAG('C','B','D','T')) &&
-          !font->GetTable(OTS_TAG('C','B','L','C'))) {
-        return OTS_FAILURE_MSG_HDR("no supported glyph shapes table(s) present");
+      if (!font->GetTable(OTS_TAG('C', 'B', 'D', 'T')) &&
+          !font->GetTable(OTS_TAG('C', 'B', 'L', 'C')) &&
+          !font->GetTable(OTS_TAG('C', 'F', 'F', '2'))) {
+        return OTS_FAILURE_MSG_HDR(
+            "no supported glyph shapes table(s) present");
       }
     }
   }

```

drott:
Fixes #160.

### DIFF
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -686,7 +686,7 @@ bool ProcessGeneric(ots::FontFile *header,
     }
   }
 
-  if (font->GetTable(OTS_TAG_CFF)) {
+  if (font->GetTable(OTS_TAG_CFF) || font->GetTable(OTS_TAG('C', 'F', 'F', '2'))) {
     // font with PostScript glyph
     if (font->version != OTS_TAG('O','T','T','O')) {
       return OTS_FAILURE_MSG_HDR("wrong font version for PostScript glyph data");


src/ots.cc
```diff
@@ -702,8 +702,9 @@ bool ProcessGeneric(ots::FontFile *header,
       // We don't sanitize bitmap tables, but don’t reject bitmap-only fonts if
       // we are asked to pass them thru.
       // Also don’t reject if we are asked to pass glyf/loca thru.
-      if (!font->GetTable(OTS_TAG('C','B','D','T')) &&
-          !font->GetTable(OTS_TAG('C','B','L','C'))) {
+      if (!font->GetTable(OTS_TAG('C', 'B', 'D', 'T')) &&
+          !font->GetTable(OTS_TAG('C', 'B', 'L', 'C')) &&
+          !font->GetTable(OTS_TAG('C', 'F', 'F', '2'))) {
```
khaledhosny: I think this check should be move up next to the `if (font->GetTable(OTS_TAG_CFF))` check.

drott: Thanks for the review, sounds reasonable. Updated the patch accordingly.

PR #162, opened 2018-03-21
Don't reject variation fonts for having a VariationIndex table

jfkthame:
As used in place of Device Tables in fonts with variation axes. This is a trivial fix to allow such a GPOS table to be parsed, rather than rejected for an invalid device table format.


khaledhosny:
Thanks!

PR #163, opened 2018-03-21
Add support for parsing GSUB/GPOS format 1.1

jfkthame:
In variation fonts, a new minor version of GSUB/GPOS may be used, to include a Feature Variations subtable. So this patch adds support for parsing that.


src/layout.cc
```diff
@@ -1496,6 +1496,160 @@ bool ParseExtensionSubtable(const Font *font,
...
+  for (uint16_t i = 0; i < substitution_count; i++) {
+    uint16_t feature_index = 0;
+    uint32_t alternate_feature_table_offset = 0;
+
+    if (!subtable.ReadU16(&feature_index) ||
+        !subtable.ReadU32(&alternate_feature_table_offset)) {
+      return OTS_FAILURE_MSG("Failed to read feature table substitution record");
+    }
+
+    if (alternate_feature_table_offset < 6 + 6 * substitution_count ||
```
khaledhosny: Should these magic numbers get names like `kFeatureTableSubstitutionHeaderSize` and `kFeatureTableSubstitutionRecordSize` for readability?

src/layout.cc
```diff
@@ -1496,6 +1496,160 @@ bool ParseExtensionSubtable(const Font *font,
...
+
+  for (uint32_t i = 0; i < feature_variation_record_count; i++) {
+    uint32_t condition_set_offset = 0;
+    uint32_t feature_table_substitution_offset = 0;
+    if (!subtable.ReadU32(&condition_set_offset) ||
+        !subtable.ReadU32(&feature_table_substitution_offset)) {
+      return OTS_FAILURE_MSG("Failed to read feature variation record");
+    }
+
+    if (condition_set_offset < subtable.offset() ||
```
khaledhosny: Shouldn’t this be `subtable.offset() + 8` to account for the header?

src/layout.cc
khaledhosny: Also, shouldn’t this be `subtable.offset() + …?`

src/layout.cc
```diff
@@ -1496,6 +1496,160 @@ bool ParseExtensionSubtable(const Font *font,
...
+        condition_set_offset > length ||
+        feature_table_substitution_offset < subtable.offset() ||
```
khaledhosny: Same here.

src/layout.cc
```diff
@@ -1496,6 +1496,160 @@ bool ParseExtensionSubtable(const Font *font,
...
+        feature_table_substitution_offset > length) {
+      return OTS_FAILURE_MSG("Offset out of range");
+    }
+
+    if (condition_set_offset) {
```
khaledhosny: Shouldn’t this offset be validated as well?

src/layout.cc
```diff
@@ -1496,6 +1496,160 @@ bool ParseExtensionSubtable(const Font *font,
...
+      if (!ParseConditionSetTable(font, data + condition_set_offset,
+                                  length - condition_set_offset)) {
+        return OTS_FAILURE_MSG("Failed to parse condition set table");
+      }
+    }
+
+    if (feature_table_substitution_offset) {
```
khaledhosny: And this one?

khaledhosny:
The ci failures should go if you rebase on master.

src/layout.cc
jfkthame: > Should these magic numbers get names...
Sure, sounds good.

> Also, shouldn’t this be `subtable.offset() + …`?
I don't think so; this is an offset from the start of the FeatureTableSubstitution table, not from the current offset within it.

jfkthame:
Rebased, fixed up offset validations, and squashed to make a tidier commit; let's see how this looks.

jfkthame:
AdobeVFPrototype (license: OFL) uses table version 1.1, so sanitization fails without the patch here (but works once it is applied).

PR #164, opened 2018-03-22
Initial stab at support for OpenType Variations tables

jfkthame:


src/fvar.cc
```diff
@@ -0,0 +1,171 @@
...
+static bool CheckTag(uint32_t tag_value) {
+  for (unsigned i = 0; i < 4; ++i) {
+    const uint32_t check = tag_value & 0xff;
+    if (check < 32 || check > 126) {
+      return false;  // non-ASCII character found.
+    }
+    tag_value >>= 8;
+  }
+  return true;
+}
```
khaledhosny: This seems to be a copy of the same function from `ots.cc`, better declare it in `ots.h` and use it here.

khaledhosny:
I think having some variable fonts for roundtrip testing would be nice, it would also help seed oss-fuzz.

jfkthame:
Adding variable fonts to tests would be good, yes. We'll need to check on licensing.... I think there are at least a couple that are OFL, so they should be OK.

brawer:
Shameless plug: For test fonts, have a look at [Unicode text rendering tests](https://github.com/unicode-org/text-rendering-tests). The [test reports](https://github.com/unicode-org/text-rendering-tests#supported-platforms) describe what’s being tested using which font.

src/fvar.cc
```diff
@@ -0,0 +1,160 @@
...
+      !table.ReadU16(&this->axisCount) ||
+      !table.ReadU16(&this->axisSize) ||
+      !table.ReadU16(&this->instanceCount) ||
+      !table.ReadU16(&this->instanceSize)) {
+    return DropVariations("Failed to read table header");
+  }
+  if (this->majorVersion != 1 /*|| this->minorVersion != 0*/) {
+    return DropVariations("Unknown table version");
+  }
+  if (this->axesArrayOffset > length || this->axesArrayOffset < table.offset()) {
```
behdad: What's wrong with a `this->axesArrayOffset < table.offset()`?

This is the kind of thing I really dislike about OTS.  Making assumptions.

src/gsub.cc
```diff
@@ -580,23 +582,34 @@ bool OpenTypeGSUB::Parse(const uint8_t *data, size_t length) {
...
+      !table.ReadU16(&version_minor) ||
       !table.ReadU16(&offset_script_list) ||
       !table.ReadU16(&offset_feature_list) ||
       !table.ReadU16(&offset_lookup_list)) {
     return Error("Incomplete table");
   }
 
-  if (version != 0x00010000) {
+  if (version_major != 1 || version_minor > 1) {
     return Error("Bad version");
```
behdad: I know this is code that has been there. Still..

Why reject font with higher minor version? That's explicitly against what the spec says.

src/fvar.cc
behdad: Yes, but what's wrong with that? An optimizing compiler can choose to do just that...  I understand it's an stretch and no one does it. But one reason I don't do those kinds of things is because of OTS.

jfkthame: Hmm... yes, in principle it could be valid. Not sure how I feel about allowing such tricksy stuff, though -- maybe it seems more likely to indicate a broken tool or corrupted font? @khaledhosny WDYT?

khaledhosny:
Can you please cleanly rebase this, it is currently unmergable due to merge conflicts.

src/fvar.cc
khaledhosny: Well, I think the idea is to protect faulty implementations from shooting themselves in the foot. I’m not sure if this particular check helps that much, but at least we should make sure all offsets + length are with in the bounds of the table they apply too.

jfkthame:
Rebased to master; I think it should now be mergeable (if builds look ok).

khaledhosny:
Thanks!

PR #165, opened 2018-04-02
Fix issues found by ots-fuzzer

jfkthame:
These are straightforward fixes: a missing offset check before calling a parsing function, and a few cases where offset/length computations using 16-bit fields might overflow (so we need to explicitly promote the values to at least 32-bit before doing arithmetic with them).


khaledhosny:
Thanks!

PR #170, opened 2018-04-27
Check the item variation store in GDEF version 1.3

jfkthame:


jfkthame:
Hmm, it seems that this causes tests/fonts/good/GradeGX4-GX-fixed-axis-name.ttf to be rejected, with the message "ERROR: GDEF: Invalid offset to glyph classes". AFAICS, this is correct: the GDEF table has version = 0x00010003, which means the header should have the additional 32-bit itemVarStoreOffset field. But that field is clearly missing; a hex dump shows the beginning of the class definition table (00 02 00 1a 00 23 00 23 ...) at the location where itemVarStoreOffset should be.

So I believe the patch here is correct, and GradeGX4-GX-fixed-axis-name.ttf should be moved to fonts/bad.

Issue #1, opened 2014-01-15
OTS rejects valide GDEF tables with ligature carets


khaledhosny:
When processing fonts having ligature carets with ot-sanitise from OTS, it gives the following warning and drops GDEF table (and subsequently GSUB and GPOS, breaking the font completely):

```
WARNING at src/gdef.cc:171: offset isn't in increasing coordinate order: 84 >= 6
ERROR at src/gdef.cc:172 (bool <anonymous namespace>::ParseLigCaretListTable(ots::OpenTypeFile *, const uint8_t *, size_t, const uint16_t))
```

However, the code referred to in the warning makes no sense at all, nothing in the [spec](http://www.microsoft.com/typography/otspec/gdef.htm) says that LigGlyph offsets have to be in coordinate order (what coordinates do LigGlyphs have anyway!), I guess it is a completely misplaced code to check for CaretValue offsets.

Issue #178, opened 2018-11-23
ERROR: GDEF: bad caret value format: 3


thundernixon:
I'm currently doing production work on a couple of variable fonts. My workflow is exporting from GlyphsApp sources to VFs, with FontMake, then checking with FontBakery.

As a concrete example, my repo of [Encode Sans](https://github.com/thundernixon/Encode-Sans) can be built by running the shell script `scripts/build.sh` in a Python3 environment with Google Fonts tools installed. As a faster example, you can view [the most recent build, TTX, and FontBakery report](https://github.com/thundernixon/Encode-Sans/tree/d1c1634c1a362187ab376011a28f28b730697c39/dist/EncodeSans-VF-2018-11-23-13_12).

I am getting the following error from OTS which I don't think is valid:

> ERROR: GDEF: bad caret value format: 3
> ERROR: GDEF: Invalid ligature caret list
> ERROR: GDEF: Failed to parse table
> Failed to sanitize file!

On line 170 of [ots/src/gdef.cc](https://github.com/khaledhosny/ots/blob/63f8d7e47cf9ab75a25b2b63bb359349fef050fa/src/gdef.cc), there is the following comment:

>       // TODO(bashi): We only support caret value format 1 and 2 for now
>       // because there are no fonts which contain caret value format 3
>       // as far as we investigated.

However, in the TTX of my variable font, I have `<CaretValue index="0" Format="3">` for each caret description in the `<LigCaretList>` table. This is also the case in [my latest Libre Caslon build](https://github.com/thundernixon/Libre-Caslon/tree/6dec2ae7747012cffd2d47aba5167c042b93e630/dist/2018-11-23-15_05).

On [MS Docs for the GDEF table](https://docs.microsoft.com/en-us/typography/opentype/otspec140/gdef#ligature-caret-list-table), it says:

> One format represents values in design units only, another fine-tunes a value based on a designated contour point, and the third uses a Device table to adjust values at specific font sizes.

I am simply following the GlyphsApp recommendation for inserting caret positions, as described on page 27 of the [Glyphs Handbook](https://glyphsapp.com/downloads/handbook/Glyphs-Handbook-2.3.pdf) - inserting an anchor labeled `caret_1`. So, I'm guessing that if my exported variable font is getting `format 3` from either GlyphsApp or from FontMake's VF exporting.

In any case, it doesn't seem to be an actual error, from what I can tell, so it may be something worth updating in OTS.

[...]

PR #182, opened 2019-01-18
Accept Ligature Caret Value Format 3

felipesanches:
(issue #178)

captura de tela 2019-01-18 as 12 57 29

See also: https://github.com/googlefonts/fontbakery/issues/2268


davelab6:
Thank you @felipesanches :)

behdad:
Funny what that piece of code does... or doesn't:
```
    // Parse caret values table 
    for (unsigned j = 0; j < caret_count; ++j) { 
      subtable.set_offset(lig_glyphs[i] + caret_value_offsets[j]); 
      uint16_t caret_format = 0; 
      if (!subtable.ReadU16(&caret_format)) { 
        return Error("Can't read caret values table %d in glyph %d", j, i); 
      } 
      // TODO(bashi): We only support caret value format 1 and 2 for now 
      // because there are no fonts which contain caret value format 3 
      // as far as we investigated. 
      if (caret_format == 0 || caret_format > kMaxCaretValueFormat) { 
        return Error("bad caret value format: %u", caret_format); 
      } 
      // CaretValueFormats contain a 2-byte field which could be 
      // arbitrary value. 
      if (!subtable.Skip(2)) { 
        return Error("Bad caret value table structure %d in glyph %d", j, i); 
      } 
    } 
```
It doesn't do anything other than checking format number.  Why do people insist on giving the false sense to font developers that ots is checking anything meaningful in their GSUB/GPOS/GDEF tables?!!

davelab6:
LOL

I guess the chrome team engineers who started OTS had their heart in the
right place :)

khaledhosny:
At least it makes sure the table do actually contains two bytes for the coordinate of each caret. Just allowing format 3 is not enough as it has more data. But since this code does not do much sensitization, one just need to check that the data is there.

src/gdef.cc
```diff
@@ -167,9 +165,6 @@ bool OpenTypeGDEF::ParseLigCaretListTable(const uint8_t *data, size_t length) {
       if (!subtable.ReadU16(&caret_format)) {
         return Error("Can't read caret values table %d in glyph %d", j, i);
       }
-      // TODO(bashi): We only support caret value format 1 and 2 for now
-      // because there are no fonts which contain caret value format 3
-      // as far as we investigated.
       if (caret_format == 0 || caret_format > kMaxCaretValueFormat) {
         return Error("bad caret value format: %u", caret_format);
       }
```
khaledhosny: You need to modify the `subtable.Skip(2)` below for format 3 as it has more data: https://docs.microsoft.com/en-us/typography/opentype/spec/gdef#caretvalue-format-3

khaledhosny:
Funnily, it was ligature carets that eventually lead me to fork OTS since it was rejecting my perfectly valid fonts because they had ligature caret. It was my very first bug fix: #1.

davelab6:
Haha! Is there a Chromium bug to render using them?

PR #188, opened 2019-03-12
Remove unused variable INNER_INDEX_BIT_COUNT_MASK

dholbert:
This variable has no usages (and doesn't seem to have had any usages in the commit that added it, https://github.com/khaledhosny/ots/commit/5ae961372b0c1709f756144f0053d0a69874b794 ).

It causes a `-Wunused-variable` build warning, so let's remove it.


dholbert:
CC @jfkthame

jfkthame:
Yep, that looks redundant -- it was there because it's defined in the spec, but we didn't end up using it for anything.

khaledhosny:
Thanks!

Issue #153, opened 2017-10-21
OTS is requiring data chunks in CMAP format 14 subtable in order?


be5invis:
As read the source it looks like that, **data chunks** of Default UVS Tables and Non-Default UVS Tables must be ordered the same as `varSelector`s, but I did not see any requirements in the Spec restricting the order of data chunks.
If the data chunks are not ordered, OTS would report an error, triggered by line 531 in cmap.cc.

khaledhosny:
PR with a test font, or just a test font, would be helpful.

be5invis:
The first file having this problem is a Microsoft IP.
The second is here ↓
[sarasa-gothic-j-regular.zip](https://github.com/khaledhosny/ots/files/1879103/sarasa-gothic-j-regular.zip)

Message:
```
ERROR: cmap: Bad subtable offset (206 != 27265)
ERROR: cmap: Failed to parse format 14 cmap subtable 2
ERROR at src\ots.cc:675 (`anonymous-namespace'::ProcessGeneric)
ERROR: cmap: Failed to parse table
Failed to sanitize file!
```

be5invis:
This message is found here: https://github.com/khaledhosny/ots/blob/master/src/cmap.cc#L531
The code forces that the last UVS data block (`DefaultUVSTable` or `NonDefaultUVSTable`) must end at the end of the `cmap` subtable, which is not required by the OTSpec.

be5invis:
@khaledhosny What exactly do you mean "something"? Something in my font or something in your program?

khaledhosny:
Something in OTS.

extratype:
@be5invis I have the same problem with Source Han Sans fonts converted to ttf using [otfcc](https://github.com/caryll/otfcc). But you have passed ots for Sarasa Gothic v0.5.4. How do you workaround this issue?

be5invis:
TTX roundtrip.

extratype:
Thank you, it works. Yesterday TTX just crashed somehow :p

khaledhosny:
Fixed in #197

PR #197, opened 2019-08-23
cmap14 table fix:

qxliu76:
This will fix the issue that: 
when both defaultUVSOffset and nonDefaultUVSOffset are non-zero and default UVS table has a larger offset value than non-default one, subtable.offset() is incorrectly set by the non default UVS table

### DIFF
--- a/src/cmap.cc
+++ b/src/cmap.cc
@@ -459,6 +459,7 @@ bool OpenTypeCMAP::Parse0514(const uint8_t *data, size_t length,
     }
   }
 
+  size_t offset = subtable.offset();
   for (unsigned i = 0; i < num_records; ++i) {
     // Checks default UVS table
     if (records[i].default_offset) {
@@ -492,6 +493,8 @@ bool OpenTypeCMAP::Parse0514(const uint8_t *data, size_t length,
         }
         last_unicode_value = check_value;
       }
+
+      offset = std::max(offset, subtable.offset());
     }
 
     // Checks non default UVS table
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks. Do you have a test font?

khaledhosny:
Yes, please. There is a test font in #153 but it is too big to include in the repo (and I’m not sure about its license either).

qxliu76:
[subset.otf.tar.gz](https://github.com/khaledhosny/ots/files/3557192/subset.otf.tar.gz)
here'the otf file.

khaledhosny:
Committed, thanks!

khaledhosny:
It seems that the `cmap` we generate from this is broken, at least it does not roundtrip with OTS:
```
$ ./ots-idempotent subset.otf
ERROR: cmap: Excessive overlap count 2
ERROR at ../src/ots.cc:693 (ProcessGeneric)
ERROR: cmap: Failed to parse table
Failed to sanitize previous output!
```

qxliu76:
> It seems that the `cmap` we generate from this is broken, at least it does not roundtrip with OTS:
> 
> ```
> $ ./ots-idempotent subset.otf
> ERROR: cmap: Excessive overlap count 2
> ERROR at ../src/ots.cc:693 (ProcessGeneric)
> ERROR: cmap: Failed to parse table
> Failed to sanitize previous output!
> ```

maybe I'm uploading the wrong file, I will check it

khaledhosny:
`ots-idempotent` sanitizes the font, then tries to sanitize the OTS output again to make sure OTS is generating fonts it thinks are valid. This second step is failing here.

PR #206, opened 2020-05-02
Remove spurious component_count check

jfkthame:
It doesn't make sense to check component_count against num_glyphs; if a component glyph is repeated within the ligature sequence, the count could legitimately be greater than the number of distinct glyphs in the font. See https://bugzilla.mozilla.org/show_bug.cgi?id=1634834.

(Also fixed a few trivial typos in messages that caught my eye.)


khaledhosny:
Makes sense, thanks (not the first time to see such bogus check against glyph count, I think I removed one or two before).

PR #207, opened 2020-05-21
Sanitize PS names more strictly than the spec requires

jfkthame:
This makes it possible for a client to override the validation of characters
in the PostScript name, to apply additional restrictions beyond what the
OpenType spec strictly requires.

This will allow clients to work around bugs in downstream software that will
consume the sanitized font data, such as Core Text; see discussion in
https://bugs.chromium.org/p/chromium/issues/detail?id=1072921.

(In https://bugzilla.mozilla.org/show_bug.cgi?id=1639813, I'm intending to
use this feature to resolve the Firefox version of that bug.)

### DIFF
--- a/src/name.cc
+++ b/src/name.cc
@@ -6,35 +6,44 @@
 
 #include <algorithm>
 #include <cstring>
+#include <cctype>
 
 // name - Naming Table
 // http://www.microsoft.com/typography/otspec/name.htm
 
 namespace {
 
-bool ValidInPsName(char c) {
-  return (c > 0x20 && c < 0x7f && !std::strchr("[](){}<>/%", c));
+// We disallow characters outside the URI spec "unreserved characters"
+// set; any chars outside this set will be replaced by underscore.
+bool AllowedInPsName(char c) {
+  return isalnum(c) || std::strchr("-._~", c);
 }
 
-bool CheckPsNameAscii(const std::string& name) {
+bool SanitizePsNameAscii(std::string& name) {
+  if (name.size() > 63)
+    return false;
+
   for (unsigned i = 0; i < name.size(); ++i) {
-    if (!ValidInPsName(name[i])) {
-      return false;
+    if (!AllowedInPsName(name[i])) {
+      name[i] = '_';
     }
   }
   return true;
 }
 
-bool CheckPsNameUtf16Be(const std::string& name) {
+bool SanitizePsNameUtf16Be(std::string& name) {
   if ((name.size() & 1) != 0)
+    return false;
+  if (name.size() > 2 * 63)
     return false;
 
   for (unsigned i = 0; i < name.size(); i += 2) {
     if (name[i] != 0) {
+      // non-Latin1 char in psname? reject it altogether
       return false;
     }
-    if (!ValidInPsName(name[i+1])) {
-      return false;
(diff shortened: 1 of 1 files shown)


khaledhosny:
I’m in favor of implementing the stricter validation directly in OTS. There is no much pointy in spec-compliance if it breaks in major implementations.

We already edit the name table, so ideally the PS name should be sanitized while not rejecting the font.

jfkthame:
> I’m in favor of implementing the stricter validation directly in OTS. There is no much pointy in spec-compliance if it breaks in major implementations.
> 
> We already edit the name table, so ideally the PS name should be sanitized while not rejecting the font.

I wondered about simply doing that, but wasn't sure we should impose it on clients that might not need/want such "weird but not strictly invalid" names to be changed. And hard-coding an ad hoc stricter validation in OTS feels a bit wrong given that we don't really know what requirements other software might have -- the chromium people happen to have discovered that "^" causes breakage in Core Text, but there could well be other stumbling-blocks (whether in Core Text or other downstream software).

It'd probably be pretty safe to enforce a really limited subset, excluding all punctuation except perhaps "-", "." and "_", but that may lead to dropping perfectly-valid PS names that clients would prefer to keep unchanged. Alternatively, we could just exclude "^", as the specific problem character that happens to have been identified now, but then we might have to revisit this when someone runs into a similar case in some other software.

(If the PS name fails validation, we don't reject the font; we already just drop it and use the built-in fallback "OTS-derived-font" instead.)

I guess I don't feel really strongly about this -- if you'd prefer to just tweak it within OTS, I can live with that although it feels a bit "wrong" to me -- please confirm which way you'd like to go. Block anything outside the safe-for-URI set?

khaledhosny:
> Block anything outside the safe-for-URI set?

I’d go for that, yes. If we can sanitize the name but not entirely drop it (e.g. replace any rejected character with `_`), that would be slightly better than the current behavior, but I wouldn’t block this issue on it. Clients should be used to the `OTS-derived-font` occasionally popping up by now.

jfkthame:
We should also make sure the name doesn't exceed the spec's length limit -- I couldn't see that we currently check this, so added it here.

jfkthame:
Just to note, I've tried a Firefox build with this patch applied and confirmed that it fixes the broken rendering of the document in https://bugzilla.mozilla.org/show_bug.cgi?id=1639813.

src/name.cc
```diff
@@ -6,35 +6,44 @@
...
-bool CheckPsNameAscii(const std::string& name) {
+bool SanitizePsNameAscii(std::string& name) {
+  if (name.size() > 63)
+    return false;
+
   for (unsigned i = 0; i < name.size(); ++i) {
-    if (!ValidInPsName(name[i])) {
-      return false;
+    if (!AllowedInPsName(name[i])) {
+      name[i] = '_';
```
khaledhosny, line 28: Just in case it is a completely invalid name (I  have seen these), we should return `false` if all characters were replaced by underscores.

jfkthame: It wouldn't really do any harm if the name ended up as "_______", would it?

khaledhosny: Of course not, it just looks strange.

khaledhosny:
Thanks!

PR #209, opened 2020-05-22
Fixup bad entry_selector in the table directory

jfkthame:
This doesn't need to be a validation failure, we can just warn and fix the value, like we do for other searchHeader fields.


khaledhosny:
Do you happen to have a test file for this?

jfkthame:
> Do you happen to have a test file for this?

No, I don't. I ran across it when looking into issues with font data that is generated dynamically to pass between processes in Firefox, but don't have an actual file. (We've now fixed the code that was sometimes generating incorrect searchHeader values, of course, but I still think we should make this change in OTS as well.)

PR #213, opened 2020-07-01
Graphite fixes

bobh0303:
Three fixes for Graphite tables:

1. correct spelling `attCollisions` -> `attrCollisions`
2. `attrCollisions` was introduced in Silf table version 4.1, not 5.0
3. The 0x800 bit is now defined and used in the FeatureDefn flags

Thanks

### DIFF
--- a/src/feat.h
+++ b/src/feat.h
@@ -34,7 +34,7 @@ class OpenTypeFEAT : public Table {
     uint32_t offset;
     uint16_t flags;
     static const uint16_t HAS_DEFAULT_SETTING = 0x4000;
-    static const uint16_t RESERVED = 0x3F00;
+    static const uint16_t RESERVED = 0x3700;
     static const uint16_t DEFAULT_SETTING = 0x00FF;
     uint16_t label;
   };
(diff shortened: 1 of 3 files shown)


bobh0303:
> Could you please squash this fixup into [770ed3f](https://github.com/khaledhosny/ots/commit/770ed3fd6a0c10b31b345f0a8bcfc033de646b73), so that we don't have an uncompilable commit in the history if possible?

I've done that in our fork, but I'm not sure how to get that pushed up to you.

jfkthame:
> > Could you please squash this fixup into [770ed3f](https://github.com/khaledhosny/ots/commit/770ed3fd6a0c10b31b345f0a8bcfc033de646b73), so that we don't have an uncompilable commit in the history if possible?
> 
> I've done that in our fork, but I'm not sure how to get that pushed up to you.

It looks like you've done so with your force-pushed update. :) This PR now just shows a single commit, which contains the (fixed-up) attrValue fix and also the new flag bit.

So, LGTM. Thanks for the fixes!

PR #214, opened 2020-07-07
[meson] split fuzzer_ldflags if isn't empty

ebraminio:
Needed for https://github.com/google/oss-fuzz/pull/4089

I followed the nice approach here for integrating oss-fuzz into the project and faced with https://github.com/google/oss-fuzz/pull/4076#discussion_r451118950 so thought should improve ots also now that it helped me :)


khaledhosny:
Thanks!

PR #216, opened 2020-09-11
Don't save variation tables which fail to parse.

garretrieger:
When parsing a variation table results in a call to DropVariations() the table is removed from m_tables by DropVariations() but not the file->tables map since it's populated later on. This means that if a TTC is being processed and a future font refers back to this table it will get a reference to the partially loaded table and result in accesses to unitialized memory.


khaledhosny:
I pushed a change to master that might make the change here unnecessary (I had it on my local copy for a few days but didn’t get over finishing and pushing it until now). Can you double check if it fixes this issue for you and then we can add only the test case.

garretrieger:
Your change fixes the problem :) I've updated this to just add the test file.

PR #220, opened 2020-11-27
meson.build: use dependencies, remove third_party/

MicroJoe:
Dear maintainer,

While trying to make a Debian package for this software in order to be able to use the tools on the distribution, I found that the meson.build was embedding the source code of the compression libraries instead of just using the "dependency" system of meson.

Embedding code is not OK in many distributions, this is why I think the third_party/ mechanism should be considered as a "last resort" way of building instead of the default. Moreover, if the source code of ots is downloaded from Gtihub releases, it cannot be built because they do not contain the source code of the third_party/ libraries.

While this PR is bold and changes everything, I hope it can start a discussion as using the third_party/ directory as a last resort instead of the default. Ideally both build mechanisms could be supported, and we would disable third_party/ build via a .patch in the Debian package for example.


MicroJoe:
To be reworked to still work with current system instead of breaking everything.

khaledhosny:
I’m fine with using meson sub-projects, but I didn’t have much success with them last time I tried. The only requirement I have is the ability to build from git clone with no additional dependencies, and to always have the same versions of the libraries.

MicroJoe:
Hello Khaled, yes, this is exactly what I though after publishing this a bit rapidly. I will come back to you with something that will:

1. Check if the dependencies can be found on the system
2. If not, use the third_party/ directory like today to build

Does this sound right?

khaledhosny:
See 306332a072a1e42cda37107b65b0e4456d53ff68 and f1e9aa6f9aa4ce49fd32fd1bc274dd702f583e87, I have to revert them because of CI build failures.

PR #222, opened 2020-11-27
meson.build: use system dependencies if possible

### LINKED ISSUE #221: OTS should be able to be built using system dependencies
This library should be able to be built from system dependencies like `libbrotli-dev`, `liblz4-dev` and `libwoff-dev` instead of only allowing to build from the copied `third_party/` submodules.

Possibility to build from the `third_party/` submodules should still be kept if the system dependencies cannot be found on the system.

(note to maintainer: creating this bug for tracking the issue in public, separate from any PR proposal)

MicroJoe:
The system dependencies should be used if available, and if not
the third_party/ dependencies are used.

Closes #221

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -37,69 +37,96 @@ endif
 configure_file(output: 'config.h',
                configuration: conf)
 
-
-brotli_c_args = []
-if 'undefined' in get_option('b_sanitize').split(',')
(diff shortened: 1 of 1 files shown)


MicroJoe:
See #223 as a better alternative for the same issue.

PR #223, opened 2020-11-28
Support external library dependencies

musicinmybrain:
Alternative resolution to #221; see my comments in that issue.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -38,68 +38,78 @@ configure_file(output: 'config.h',
                configuration: conf)
 
 
-brotli_c_args = []
-if 'undefined' in get_option('b_sanitize').split(',')
-  # brotli depends on undefined behavior otherwise.
-  brotli_c_args += '-DBROTLI_BUILD_PORTABLE'
+ots_libs = []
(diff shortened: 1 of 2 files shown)


khaledhosny:
Meson already have built-in support for building dependencies from source if they are not found, so that should be used instead of adding custom options, and distributions already know how to handle it and disable any downloads: https://mesonbuild.com/Subprojects.html

khaledhosny (changes requested):
Meson already have built-in support for building dependencies from source if they are not found, so that should be used instead of adding custom options, and distributions already know how to handle it and disable any downloads: https://mesonbuild.com/Subprojects.html

musicinmybrain:
Does that mean your preferred approach is to migrate the woff2/brotli and lz4 dependencies from git submodules to meson subprojects? (An obvious prerequisite would be adding woff2 to the wrapdb.)

khaledhosny:
Yes, if possible.

musicinmybrain:
All right; thanks for the feedback. I will work on a new PR.

PR #225, opened 2020-11-29
introduce man page ots-sanitize.1

MicroJoe:


docs/ots-sanitize.1
```diff
@@ -0,0 +1,51 @@
...
+.PP
+ots-sanitize is a program which validates and/or transcodes a font file using
+the OTS library:
+.PP
+.RS
+.nf
+transcoded_font = ValidateAndTranscode(original_font);
+if (validation_error)
+  PrintErrorAndExit;
+OutputToStdout(transcoded_font);
```
khaledhosny, line 18: Not sure why this pseudo code is here, without even a mention what it is, but I find it strange for man page.

khaledhosny (changes requested):

khaledhosny:
If we are providing a man page, then we should install it as well.

docs/ots-sanitize.1
MicroJoe: This was copy-pasted from `DesignDoc.md`, removed in new patchset.

MicroJoe:
> If we are providing a man page, then we should install it as well.

Done, installation is done via new instruction in meson.build.

khaledhosny (approved):

PR #227, opened 2020-12-04
Sanitise values for fonts with invalid maxPoints and maxComponentPoints

drott:
Sanitize values in fonts that exceed values maxPoints
and maxComponentPoints from the maxp table.

While traversing components, also keep track of depth vs.
maxComponentDepth and overwrite in maxp when a larger value is needed.

Reject only when values exceed 0xFFFF as that is the maximum value
for the limits in maxp.

Add test fonts to test_good_fonts, based on Amiri font subsets:
01ae09f3a2ca8f33035e6261d09e9fe06b919174.ttf  Max composite levels 3, instead of 4
ad455c8c531334cf470a98901cbe86a378895c14.ttf  Max composite points 68 instead of 69
49331b1dd031e95ca803d632f69404d09ee6f592.ttf  Max points 40 instead of 41


jfkthame:
Do we actually want to reject these, or should OTS just fix the maxp table to have the correct values? @khaledhosny WDYT?

drott:
> Do we actually want to reject these, or should OTS just fix the maxp table to have the correct values? @khaledhosny WDYT?

Sorry, the commit message is wrong, I'll update it. I actually went to change it to fixup the maxp values.

jfkthame:
> > Do we actually want to reject these, or should OTS just fix the maxp table to have the correct values? @khaledhosny WDYT?
> 
> Sorry, the commit message is wrong, I'll update it. I actually went to change it to fixup the maxp values.

Aha! Sounds good, thanks. (As you can tell, I didn't actually look at the code, just saw the message.)

jfkthame:
Maybe it'd be nice to keep track of these maxima across all the glyphs, and then adjust the values in maxp at the end, so that we can correctly *reduce* them if they were larger than necessary, as well as increase them if needed? It'd mean tracking them separately from the values in maxp while processing the glyphs, instead of just adjusting maxp on the fly, but it seems like it shouldn't be much harder.

drott:
> Maybe it'd be nice to keep track of these maxima across all the glyphs, and then adjust the values in maxp at the end, so that we can correctly _reduce_ them if they were larger than necessary

Thanks for the suggestion, I have an experimental implementation of this in https://github.com/drott/ots/commit/6925e6ba380c8ca8bd4b7adc8a27e4c5ba4462c6 - but I'd prefer to do this separately. Due to a [bug in FontTools](https://github.com/fonttools/fonttools/pull/2045) computing too large values for `maxComponentDepth` and even some stock Windows 10 fonts having too large component depth values, I would consider that change more experimental and my priority is on raising them. Fonts with too large values would not be considered invalid.

src/glyf.cc
```diff
@@ -302,9 +313,68 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
...
+      ComponentPointCount component_point_count;
+      component_point_count.accumulated_component_points = 0;
+      component_point_count.gid_stack.clear();
+
+      if (!ParseCompositeGlyph(glyph, &component_point_count)) {
         return Error("Failed to parse glyph %d", i);
       }
+
+      // Check maxComponentDepth and validate
+      // maxComponentPoints. ParseCompositeGlyph placed the first set of
```
khaledhosny, line 326: The line break here is weird.

src/glyf.cc
```diff
@@ -302,9 +313,68 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
...
+                       length);
+        }
+        if (gly_offset + gly_length < gly_offset) {
+          return Error("Glyph %d length (%d < 0)!", i, gly_length);
+        }
+        if (gly_offset + gly_length > length) {
+          return Error("Glyph %d length %d too high", i, gly_length);
+        }
+
+        Buffer points_count_glyph(data + gly_offset, gly_length);
```
khaledhosny, line 354: I feel uneasy about the duplicated code (from `gly_offset` to here, may be it should be factored out (`GetGlyphData()` or something)?

src/glyf.cc
```diff
@@ -344,6 +414,88 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
...
+      if (!glyph.ReadU16(&tmp_index)) {
+        return Error("Can't read contour index %d", i);
+      }
+      num_points = tmp_index + 1;
+    }
+
+    component_point_count->accumulated_component_points += num_points;
+    return true;
+  } else  {
+    assert(num_contours == -1);
```
khaledhosny, line 464: Since we are reading `num_contours` from input data, I don’t think assert is good here. See the checks for `num_contours` in `OpenTypeGLYF::Parse()` (again, I feel uneasy about the code duplication).

src/glyf.h
```diff
@@ -23,12 +23,31 @@ class OpenTypeGLYF : public Table {
...
 
  private:
+  struct GidAtLevel {
+    uint16_t gid;
+    uint32_t level;
+  };
+
+  struct ComponentPointCount {
+    uint32_t accumulated_component_points;
+    std::vector<GidAtLevel> gid_stack;
```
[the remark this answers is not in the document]
drott: Google C++ style guide suggests to use named members of structs for readability:
https://google.github.io/styleguide/cppguide.html#Structs_vs._Tuples

src/glyf.cc
```diff
@@ -302,9 +313,68 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
         return Error("Failed to parse glyph %d", i);
       }
     } else {
-      if (!ParseCompositeGlyph(glyph)) {
+
+      ComponentPointCount component_point_count;
+      component_point_count.accumulated_component_points = 0;
```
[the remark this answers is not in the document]
drott: Initialisation of ComponentPointCount fixed according to comments.

src/glyf.cc
```diff
@@ -302,9 +313,68 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
...
+      component_point_count.gid_stack.clear();
```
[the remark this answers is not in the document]
drott: Removed.

src/glyf.cc
drott: Formatting updated.

src/glyf.cc
drott: Factored out glyph buffer retrieval out into new function `GetGlyphBufferSection()`.

src/glyf.cc
drott: All previous conditions, `num_contours <= -2`, `0`, `> 0` have been checked, that's why I was using the `assert` here for clarity. Agree that the duplication would be preferably avoided, I have been thinking about that as well. I believe it would mean some kind of recursive call to a paramaterized Parse(), but Parse() has many side effects such as adding to the iov output buffer, which are hard to encapsulate. Also, the intention of this traversal is really only focusing on a fast points count, not on additional validations such as hinting instruction count or logic checks on xmin, xmax etc. coordinates - hence I was considering it acceptable to repeat, also given that the `glyf` table structure is unlikely to change. 

Alternatively, I have been whether it'd make sense to keep some kind of parallel vector of points count in the iterative top level traversal in parse, and pushing points count for components upwards in that, but this does not easily work out as a glyf can be component to multiple other glyfs and there is no requirement of a topological order, so multiple passes would be required IIUC. Also, the idea of creating copies seemed wasteful, as we're anyway traversing over an existing font blob in memory.

khaledhosny: > All previous conditions, `num_contours <= -2`, `0`, `> 0` have been checked

That is true for the glyph where `TraverseComponentsCountingPoints()` , but is it true for the glyphs that `TraverseComponentsCountingPoints()` is processing? e.g. suppose `A` is a composite glyph of `A.1` and `A.2`, by the time `TraverseComponentsCountingPoints()` is called on `A.1` or r `A.2` they may or may not have been sanitized by `Parse()`, or am I confusing myself?

drott: Bear with me if I missed something - I am trying to understand what your direction. IIUC there are two things to discuss here:

- I am using the assert() only to document that num_contours can only be `-1` at this point, at the same code nesting level, previous if conditions have checked for all other values.
- The other thing which you raise here: Order in which all glyphs get sanitized: `OpenTypeGLYF::Parse` iterates glyph by glyph, in the glyph id order. If the glyph has components, at the end of the loop body, this change here starts traversing the components with the intention of computing adherence to maxComponentDepth and maxComponentPoints and adjusting `maxp` when needed. Going off on this traversal, things not related to this traversal are skipped as much as possible, e.g. xmin/xmax validation as well as hinting instructions related things - as they don't affect points count. But each glyph component itself will eventually be visited by the glyph loop in Parse and checks for xmin, max, and TT instructions will be performed.

khaledhosny: OK, sorry for the confusion. I missed the check for `num_contours <= -2` in `TraverseComponentsCountingPoints()` and thought only `Parse()` checked for it. I’m still uneasy about the duplicate checks, but I doubt it will be easy to miss if in the future `-2` etc. started to have a meaning as the code will error loudly.

khaledhosny (approved):
LGM. If @jfkthame is happy with this I’ll merge.

drott:
Thank you, I appreciate the review.

PR #230, opened 2020-12-15
Default initialize maxp members

drott:
Should fix
https://bugs.chromium.org/p/chromium/issues/detail?id=1158774

### DIFF
--- a/src/maxp.h
+++ b/src/maxp.h
@@ -17,24 +17,24 @@ class OpenTypeMAXP : public Table {
   bool Parse(const uint8_t *data, size_t length);
   bool Serialize(OTSStream *out);
 
-  uint16_t num_glyphs;
-  bool version_1;
+  uint16_t num_glyphs = 0;
+  bool version_1 = 0;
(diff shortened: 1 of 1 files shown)


khaledhosny:
This is already fixed in 1141c81c411b599e40496679129d0884715e8650. I didn’t do the initialization (though it is safer) as I’d rather the let fuzzer catch this. A proper fix is to reject fonts with maxp version 0 and glyf table or maxp 1 and CFF table.

khaledhosny:
(I can’t see the chromium issue, but I assume it is the same uninitialized memory issue found by oss-fuzz earlier).

drott:
Good, thanks for the explanation, works for me.

drott:
> (I can’t see the chromium issue, but I assume it is the same uninitialized memory issue found by oss-fuzz earlier).

Yes, probably similar. I had one CL where I only used the new maxp sanitization without your follow-up change so that the Chromium fuzzer found this, too.

PR #231, opened 2020-12-16
Rewrite manual page in mdoc(7)

Alhadis:
> From a [suggestion](https://github.com/khaledhosny/ots/commit/d4e981df41efae4d1016893f1a192430c406a220#r45167406) I left on d4e981d, folded into a PR as per @khaledhosny's request.

## Description
The changes applied by this pull-request replace the use of legacy [`man(7)`][man] macros with a modern alternative: [`mdoc(7)`][mdoc].

Practical benefits include:
1. Consistent output, both in structure and formatting
2. Semantic, self-documenting markup
3. Clearer syntax. E.g., authors need not write `-` as `\-` to output ASCII hyphen-minus (U+002D) instead of Unicode hyphen (U+2010).

## Preview

Figure 1

The files used to generate the above frames are attached in pre-formatted (“catman”) form:
- **Before:** [man.txt](https://github.com/khaledhosny/ots/files/5702323/man.txt)
- **After:** [mdoc.txt](https://github.com/khaledhosny/ots/files/5702330/mdoc.txt)

[man]: http://man.openbsd.org/man.7
[mdoc]: http://man.openbsd.org/mdoc.7

### DIFF
--- a/docs/ots-sanitize.1
+++ b/docs/ots-sanitize.1
@@ -1,42 +1,49 @@
-.TH OTS-SANITIZE "1" "November 2020" "OpenType Sanitizer" "User Commands"
-.SH NAME
-ots-sanitize \- font validator and transcoder
-.SH SYNOPSIS
-.B yes
-[\fI\,OPTIONS\/\fR]... \fI\,FONT_FILE\/\fR [\fI\,DEST_FONT_FILE\/\fR] [\fI\,FONT_INDEX\/\fR]
-.SH DESCRIPTION
-.\" Add any additional description here
-.PP
-ots-sanitize is a program which validates and/or transcodes a font file using
-the OTS library.
-.PP
-The OpenType Sanitizer (OTS) parses and serializes OpenType files (OTF, TTF)
+.Dd November 2020
+.Dt OTS-SANITIZE 1
+.Os
+.Sh NAME
+.Nm ots-sanitize
+.Nd font validator and transcoder
+.
+.Sh SYNOPSIS
+.Nm
+.	Op Fl -quiet
+.	Ar    FONT_FILE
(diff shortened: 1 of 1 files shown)


Alhadis:
@khaledhosny Truth be known, there's a lot more that could be added to this page, but I restrained myself from going out-of-scope. In particular, the synopses documented in [`HowToTest.md`](https://github.com/khaledhosny/ots/blob/master/docs/HowToTest.md#command-line-tools) could be converted to `mdoc(7)` using some massaging with `awk(1)` or `sed(1)`.

To learn what other mdoc tags are available, I recommend [OpenBSD's version of mdoc(7)](http://man.openbsd.org/mdoc.7).

khaledhosny:
I have no use for the man page, since the current one was contributed by @MicroJoe, I’ll leave to to him to decide if the changes here are fine for him or not.

MicroJoe:
Hello, I do not understand why "BSD General Commands Manual" should replace "User Commands", and "BSD" replace "OpenType Sanitizer" in the visual diff.

This is probably from the new formatting language that is assuming everything is BSD by default. Maybe this can be fixed?

Alhadis:
@MicroJoe That's just a [standard header](https://github.com/unofficial-mirror/groff/blob/925c0cc9fe98e102e4e4ae501acfeeee002713ac/tmac/mdoc/doc-common-u#L239) that's the same for every manual page that doesn't specify an operating system explicitly (via the [`.Os`](http://man.openbsd.org/mdoc.7#Os) macro). The recommended practice is to leave it blank, which uses the site default (on Darwin/macOS, "BSD" is displayed presumably because it itself is a BSD-derived system).

In any case, the matter of BSD's section headers is unrelated to this PR.

PR #234, opened 2021-05-11
Add man pages for installed binaries that lack them

musicinmybrain:
These use the same style as the existing `ots-sanitize.1`. Like that man page, they are based directly on the contents of `HowToTest.md`.

The new man pages are added to `meson.build`, and all man pages are cross-referenced in a new `SEE ALSO` section.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -160,6 +160,7 @@ executable('ots-sanitize',
   link_with: libots,
   install: true,
 )
+install_man('docs/ots-sanitize.1')
 
 
 fuzzer_ldflags = []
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -239,7 +239,13 @@ if cxx.has_header('sys/time.h')
...
 
 
-install_man('docs/ots-sanitize.1')
+install_man([
+  'docs/ots-idempotent.1',
+  'docs/ots-perf.1',
+  'docs/ots-sanitize.1',
+  'docs/ots-side-by-side.1',
+  'docs/ots-validator-checker.1',
+])
```
khaledhosny, line 248: Some of these tools are built conditionally, so man pages should only be installed when the tools are installed.

khaledhosny (changes requested):

meson.build
musicinmybrain: Thanks; I‘ll look at the details and revise.

khaledhosny (approved):

PR #235, opened 2021-06-14
Warn if glyph name too long

simoncozens:
According to the [post table spec](https://docs.microsoft.com/en-us/typography/opentype/spec/post),

> Names must be no longer than 63 characters; some older implementations may assume a length limit of 31 characters.

This implements a check for name length of glyph names in the post table. As glyph names are purely informational, I have made this a warning rather than an error.

### DIFF
--- a/src/post.cc
+++ b/src/post.cc
@@ -11,6 +11,10 @@
 
 namespace ots {
 
+// "Names must be no longer than 63 characters; some older implementations may
+// assume a length limit of 31 characters."
+const int MAX_GLYPH_NAME_LENGTH = 63;
+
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Thanks Simon! Yeah, this will break poorly implemented software, like FontForge. (Cf. fontforge/fontforge#4751.) I imagine other similarly badly-thought-out C code is just doing `char[63]` somewhere.

khaledhosny:
We either should truncate the names (tricky as it can result in duplicate names) or drop the names entirely and change the table version to 3 (should be safe on the web), if we really think this is potentially an issue. Warnings are good, but OTS main job is a sanitizer not a linting tool.

PR #238, opened 2021-12-28
Fix CFF2 vsindex handling

josh-hadley:
This fixes several issues with the previous handling of vsindex in CFF2 CharStrings:
 1) vsindex value, when set to a value other than the default in a CharString, now correctly persists through calls to subroutines (which may in turn call blend, whose argument parsing depends on the vsindex value).
 2) Per the CFF2 specification, the initial value for vsindex for a CharString comes from the PrivateDICT associated with the glyph, rather than assuming a default of 0.

Summary of changes:
- added CharStringContext class to hold various values that need to persist across subr calls within CharStrings
- pass the above context in to nested subroutines so they have access to the correct values
- add a mechanism for retrieving the vsindex value of individual FontDICT->PrivateDICT values so that vsindex is initially set for each CharString according to the CFF2 specification
- add test font for vsindex/blend in subrs
- update some outdated comments


khaledhosny (approved):
LGTM. Please add the new test file to the list in `tests/meson.build`.

khaledhosny:
This fuzzed file causes an abort because of unhanded `std::out_of_range` exception. Seems like the CFF2 table has no private dict so `vsindex_per_font` never gets populated. We probably need to make sure it gets populated, and also do a bounds check when accessing and reject the table in this case.
[clusterfuzz-testcase-minimized-ots-fuzzer-4532447398723584.zip](https://github.com/khaledhosny/ots/files/7790372/clusterfuzz-testcase-minimized-ots-fuzzer-4532447398723584.zip)

josh-hadley:
Indeed -- I'll work on an update to handle this case a bit better, as you describe. Thanks for the extra check!

khaledhosny:
Thanks to OSS-Fuzz for ever coming up with more ways to generate broken fonts :)

josh-hadley:
Shall I add that file (renamed with sha) to the fuzzing folder & test set?

khaledhosny:
Yes, please.

PR #239, opened 2021-12-30
Avoid unhanded std::out_of_range exception in bad CFF2 font

josh-hadley:
As discussed at https://github.com/khaledhosny/ots/pull/238#issuecomment-1002750392.
Note that the root cause is that there is *no* PrivateDICT at all (not even an empty one), which is required for CFF2, and that condition was not being checked. So I added a check for that and fail early. In case we still somehow get to ValidateCFFCharStrings and attempt to access vsindex_per_font, I also added a check there to avoid an out-of-range exception. Also updated test suite with the font in question.


src/cff.cc
```diff
@@ -530,6 +534,7 @@ bool ParsePrivateDictData(
         if (vsindex >= out_cff->region_index_count.size()) {
           return OTS_FAILURE();
         }
+        out_cff->vsindex_per_font.back() = vsindex;
```
khaledhosny, line 537: The scope of `vsindex` can be reduced now by moving the declaration to where it is set

khaledhosny (approved):

src/cff.cc
josh-hadley: `vsindex` is also used when parsing `blend`, just below that block. Would you prefer changing the `blend` case to use `out_cff->vsindex_per_font.back()` in order to reduce the scope of `vsindex` as you propose? Or reduce the scope to the [`while` block](https://github.com/khaledhosny/ots/pull/239/files#diff-8a7f0692284ab2a9a55e74c3218713bbfb02005e48199f2a73aa9f15b3263b65R420) level?

khaledhosny: I see. Let's leave it the way it is for now.

PR #244, opened 2022-07-06
Exit early if decompressed size exceeds OTSStream's size limit

fred-wang:
Currently we are exiting early if decompressed size exceeds
OTS_MAX_DECOMPRESSED_FILE_SIZE=300MB but some callers may actually set a
smaller limit. For example, Chromium uses 30MB [1] and its internal
fuzzer 256kB [2]. Firefox uses a limit of 256MB [3].

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/fonts/web_font_decoder.cc;l=155;drc=5863e513ede6d1f7e9a060f13635fd3916a2183c
[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/ots/fuzz/ots_fuzzer.cc;l=11;drc=efffce8e2a6da2d2295472ed24ef040e0bf4ea14
[3] https://searchfox.org/mozilla-central/rev/3e1a721bce1da3ae04675539b39a4e95b25a046d/gfx/thebes/gfxOTSUtils.h#28

### DIFF
--- a/include/opentype-sanitiser.h
+++ b/include/opentype-sanitiser.h
@@ -58,7 +58,7 @@ class OTSStream {
 
   virtual ~OTSStream() {}
 
-  virtual size_t SizeLimit() = 0;
+  virtual size_t size() = 0;
 
   // This should be implemented to perform the actual write.
   virtual bool WriteRaw(const void *data, size_t length) = 0;
--- a/include/opentype-sanitiser.h
+++ b/include/opentype-sanitiser.h
@@ -58,6 +58,8 @@ class OTSStream {
 
   virtual ~OTSStream() {}
 
+  virtual size_t SizeLimit() = 0;
+
   // This should be implemented to perform the actual write.
   virtual bool WriteRaw(const void *data, size_t length) = 0;
 
--- a/include/ots-memory-stream.h
+++ b/include/ots-memory-stream.h
@@ -18,7 +18,7 @@ class MemoryStream : public OTSStream {
       : ptr_(ptr), length_(length), off_(0) {
   }
 
-  virtual size_t SizeLimit() { return length_; }
+  size_t size() override { return length_; }
 
   virtual bool WriteRaw(const void *data, size_t length) {
     if ((off_ + length > length_) ||
@@ -62,7 +62,7 @@ class ExpandingMemoryStream : public OTSStream {
     return ptr_;
   }
 
-  size_t SizeLimit() { return limit_; }
+  size_t size() override { return limit_; }
 
   bool WriteRaw(const void *data, size_t length) {
     if ((off_ + length > length_) ||
--- a/include/ots-memory-stream.h
+++ b/include/ots-memory-stream.h
@@ -18,6 +18,8 @@ class MemoryStream : public OTSStream {
       : ptr_(ptr), length_(length), off_(0) {
   }
 
+  virtual size_t SizeLimit() { return length_; }
+
   virtual bool WriteRaw(const void *data, size_t length) {
     if ((off_ + length > length_) ||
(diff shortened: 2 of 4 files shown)


fred-wang:
cc @khaledhosny @jfkthame @drott

fred-wang:
@jfkthame gfxOTSExpandingMemoryStream will need a 

          size_t SizeLimit() override { return m_Limit; }

if this is taken.

include/opentype-sanitiser.h
```diff
@@ -58,6 +58,8 @@ class OTSStream {
 
   virtual ~OTSStream() {}
 
+  virtual size_t SizeLimit() = 0;
```
khaledhosny, line 61: I’d rename it `size()`.

src/ots.cc
```diff
@@ -531,6 +531,10 @@ bool ProcessWOFF2(ots::FontFile *header,
                                OTS_MAX_DECOMPRESSED_FILE_SIZE / (1024.0 * 1024.0));
   }
 
+  if (decompressed_size > output->SizeLimit()) {
+    return OTS_FAILURE_MSG_HDR("Size of decompressed WOFF 2.0 font exceeds OTSStream's size limit (%gMB)", output->SizeLimit() / (1024.0 * 1024.0));
```
khaledhosny, line 535:
```suggestion
    return OTS_FAILURE_MSG_HDR("Size of decompressed WOFF 2.0 font exceeds output size (%gMB)", output->SizeLimit() / (1024.0 * 1024.0));
```

src/ots.cc
```diff
@@ -672,6 +676,10 @@ bool ProcessGeneric(ots::FontFile *header,
                                OTS_MAX_DECOMPRESSED_FILE_SIZE / (1024.0 * 1024.0));        
   }
 
+  if (uncompressed_sum > output->SizeLimit()) {
+    return OTS_FAILURE_MSG_HDR("decompressed sum exceeds OTSStream's size limit (%gMB)", output->SizeLimit() / (1024.0 * 1024.0));
```
khaledhosny, line 680:
```suggestion
    return OTS_FAILURE_MSG_HDR("decompressed sum exceeds output size (%gMB)", output->SizeLimit() / (1024.0 * 1024.0));
```

include/ots-memory-stream.h
```diff
@@ -60,6 +62,8 @@ class ExpandingMemoryStream : public OTSStream {
     return ptr_;
   }
 
+  virtual size_t SizeLimit() { return limit_; }
```
khaledhosny, line 65:
```suggestion
  size_t SizeLimit() { return limit_; }
```

khaledhosny (approved):
On a second thought, it makes sense.

include/opentype-sanitiser.h
fred-wang: Done. Existing methods in OTSStream seemed to use CamelCase but I see that the trival one chksum() does not.

include/ots-memory-stream.h
fred-wang: I was doing the same as WriteRaw which repeats the "virtual" keyword in the derived classes. But I don't like it either, so instead I replaced that with "override" in the derived classes.

fred-wang:
mmh, I'm getting this kind of warnings when trying to port to chromium, so I guess we should use override everywhere

../../third_party/ots/src/include/ots-memory-stream.h:23:16: error: 'WriteRaw' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  virtual bool WriteRaw(const void *data, size_t length) {
               ^

PR #245, opened 2022-07-08
Use override keyword when implementing virtual members of OTSStream

fred-wang:
See https://github.com/khaledhosny/ots/pull/244#issuecomment-1178914779


khaledhosny:
Thanks

PR #247, opened 2022-09-05
Validate the color-font CPAL and COLR tables

jfkthame:
This adds validation for the CPAL and COLR (v.0 and v.1) tables. It'd be good to have support for these in ots-sanitize, for the benefit of font designers / tool developers. In addition, it can provide a first line of defense for browser implementations that are exposed to potentially damaged or malicious webfont resources.

@khaledhosny any comments/corrections much appreciated -- thanks!


src/cpal.cc
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  if (paletteEntryLabelsArrayOffset) {
+    if (paletteEntryLabelsArrayOffset < headerSize || paletteEntryLabelsArrayOffset >= length) {
+      return Error("Bad palette entry labels array offset in table header");
+    }
+    if (!ParsePaletteEntryLabelsArray(font, data + paletteEntryLabelsArrayOffset, length - paletteEntryLabelsArrayOffset,
+                                      numPaletteEntries)) {
+      return Error("Failed to parse palette labels array");
```
jfkthame, line 178: s/palette labels/palette entry labels/

src/cpal.cc
```diff
@@ -0,0 +1,200 @@
...
+      return Error("Failed to read CPAL table header");
+    }
+    headerSize += 3 * sizeof(uint32_t);
+  }
+
+  if (colorRecordsArrayOffset < headerSize || colorRecordsArrayOffset >= length) {
+    return Error("Bad color records array offset in table header");
+    if (colorRecordsArrayOffset > length - size_t(numColorRecords) * 4) {
+      return Error("Color records array exceeds table bounds");
+    }
```
jfkthame, line 149: Outside the other if!

khaledhosny:
Thanks Jonathan.

I haven’t reviewed the code yet, but here is a malicious font I created by accident a while ago that crashed two implementations at that time, https://bugs.chromium.org/p/chromium/issues/detail?id=1274206, and it does not seem like the code here does anything to it.

jfkthame:
> Thanks Jonathan.
> 
> I haven’t reviewed the code yet, but here is a malicious font I created by accident a while ago that crashed two implementations at that time, https://bugs.chromium.org/p/chromium/issues/detail?id=1274206, and it does not seem like the code here does anything to it.

So according to that issue, "one of the color glyphs refers to itself", i.e. the graph of paint records has a cycle (of the most trivial kind). Yes, I ran into that during development of the code in Gecko, too. :-)

At this point, I explicitly did *not* check for cycles in the OTS code, because the [COLR table spec](https://docs.microsoft.com/en-gb/typography/opentype/spec/colr#color-glyphs-as-a-directed-acyclic-graph) explicitly tells applications to check for this, and to terminate rendering when a cycle is detected (but the graph up to that point should be rendered):

> Applications should track paint tables within a path in the graph, checking whether any paint table was already encountered within that path. [...] If the graph contains a cycle or is otherwise not well formed or valid, the paint table at which the error occurs should be ignored, that sub-graph should not be rendered, and that node in the graph should be considered to be visually bounded. The application should attempt to render the remainder of the graph, if well formed and valid.

So I didn't feel it was right to reject a font entirely due to a cycle. But perhaps we should, given that if a rendering implementation isn't robust against this, it's a trivial way to send it into an infinite loop/recursion?

khaledhosny:
> But perhaps we should, given that if a rendering implementation isn't robust against this, it's a trivial way to send it into an infinite loop/recursion?

That would be my inclination as well.

jfkthame:
> > But perhaps we should, given that if a rendering implementation isn't robust against this, it's a trivial way to send it into an infinite loop/recursion?
> 
> That would be my inclination as well.

Fair enough; I'll add a check for that.

(I did just try the example from that Chromium report in Firefox Nightly, and confirmed that we render it safely -- phew!)

Interestingly, in Chrome Canary, it doesn't seem to render anything at all for me, as if they're rejecting the glyph entirely rather than rendering up to the point where the cycle occurs. But that's OK; my reading of the spec is that allows for that option:

> If an application is not able to recover from errors while traversing the graph, it may ignore the color glyph entirely.

jfkthame:
I've revised the code to include checking for cycles; it now rejects your "accidentally malicious" example.

src/cpal.cc
```diff
@@ -0,0 +1,199 @@
...
+
+  uint32_t headerSize = 4 * sizeof(uint16_t) + sizeof(uint32_t) +
+      numPalettes * sizeof(uint16_t);
+
+  // uint16_t colorRecordIndices[numPalettes]
+  for (auto i = 0u; i < numPalettes; ++i) {
+    uint16_t colorRecordIndex;
+    if (!table.ReadU16(&colorRecordIndex)) {
+      return Error("Failed to read color record index");
+    }
```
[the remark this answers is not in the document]
jfkthame: Good point, will fix.

src/cpal.cc
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  for (auto i = 0u; i < numPaletteEntries; ++i) {
+    uint16_t nameID;
+    if (!subtable.ReadU16(&nameID)) {
+      return OTS_FAILURE_MSG("Failed to read palette entry label ID for index %u", i);
+    }
+    if (!name->IsValidNameId(nameID)) {
+      // Should we treat this as a failure?
+      OTS_WARNING("Palette entry %u label ID %u missing from name table", i, nameID);
```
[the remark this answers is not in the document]
jfkthame: Checking the spec, it looks like this applies to both palette labels and palette entry labels, actually. I'll see what I can do about this - thanks!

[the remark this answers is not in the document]
jfkthame: I've updated the code to de-serialize/serialize, as suggested, and to fix up warning-level issues found (missing name IDs are replaced with 0xFFFF, and reserved paletteType flag bits are cleared).

src/colr.cc
```diff
@@ -0,0 +1,1069 @@
...
+  return true;
+}
+
+bool ParsePaintLinearGradient(const ots::Font* font,
+                              const uint8_t* data, size_t length,
+                              colrState& state, bool var)
+{
+  ots::Buffer subtable(data, length);
+
+  uint32_t colorLine;
```
khaledhosny, line 266: This should be `colorLineOffset` to make it clear it is an offset.

src/colr.cc
```diff
@@ -0,0 +1,1069 @@
...
+
+bool ParsePaintComposite(const ots::Font* font,
+                         const uint8_t* data, size_t length,
+                         colrState& state)
+{
+  ots::Buffer subtable(data, length);
+
+  uint32_t sourcePaint;
+  uint8_t compositeMode;
+  uint32_t backdropPaint;
```
khaledhosny, line 591: Should be `sourcePaintOffset` and `backdropPaintOffset`.

khaledhosny (approved):
We need some test fonts and addressing the minor nits above, but LGTM otherwise.

src/colr.cc
```diff
@@ -0,0 +1,1069 @@
...
+  if (!table.ReadU16(&version) ||
+      !table.ReadU16(&numBaseGlyphRecords) ||
+      !table.ReadU32(&offsetBaseGlyphRecords) ||
+      !table.ReadU32(&offsetLayerRecords) ||
+      !table.ReadU16(&numLayerRecords)) {
+    return Error("Incomplete table");
+  }
+
+  if (version > 1) {
+    return Error("Bad version");
```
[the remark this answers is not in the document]
jfkthame: Done.

src/colr.cc
```diff
@@ -0,0 +1,1069 @@
...
+
+  if (alpha < 0 || alpha > F2DOT14_one) {
+    OTS_WARNING("Alpha value outside valid range 0.0 - 1.0");
+  }
+
+  if (var) {
+    VarIdxBase varIndexBase;
+    if (!subtable.ReadU32(&varIndexBase)) {
+      return OTS_FAILURE_MSG("Failed to read PaintVarSolid");
+    }
```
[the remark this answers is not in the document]
jfkthame: I don't think we can, because it's perfectly legal to use an index that is outside the range of the DeltaSetIndexMap; [the spec says](https://docs.microsoft.com/en-gb/typography/opentype/spec/otvarcommonformats#associating-target-items-to-variation-data):

> In the context in which a DeltaSetIndexMap table is used, if an index into the mapping array is used that is greater than or equal to mapCount, then the last logical entry of the mapping array is used.

src/colr.cc
jfkthame: Done.

src/colr.cc
jfkthame: Done.

jfkthame:
@khaledhosny Thanks for the review, much appreciated!

I'm not sure what your usual practice is for adding test fonts? For some examples that include COLR/CPAL tables, I'd suggest we can pull the `*colr_v1.otf` and `*colr_v1.ttf` files from https://github.com/googlefonts/color-fonts/tree/main/fonts. Should I simply grab these and add them (with SHA-1 names) to `tests/fonts/good`?

What about fuzzing inputs? I see there's a `tests/fonts/fuzzing` directory, but I'm unclear what the expectations are there. If I simply drop some COLR/CPAL fonts in there, will the fuzzing architecture automatically pick them up and start munging them?

khaledhosny:
> @khaledhosny Thanks for the review, much appreciated!
> 
> I'm not sure what your usual practice is for adding test fonts? For some examples that include COLR/CPAL tables, I'd suggest we can pull the `*colr_1.otf` and `*colr_1.ttf` files from https://github.com/googlefonts/color-fonts/tree/main/fonts. Should I simply grab these and add them (with SHA-1 names) to `tests/fonts/good`?

Yes, that should do it. We can also add the broken font in the bad directory.

> What about fuzzing inputs? I see there's a `tests/fonts/fuzzing` directory, but I'm unclear what the expectations are there. If I simply drop some COLR/CPAL fonts in there, will the fuzzing architecture automatically pick them up and start munging them?

All fonts are fed to the fuzzer. These ones are from crash test cases to make sure we don’t regress and re-introduce the crash (we don’t much care if they pass or not, just that they don’t cause a crash).

jfkthame:
Thanks! I was just starting to find my way around the test setup... will add a commit with fonts shortly.

jfkthame:
@khaledhosny Unless you see anything else, I think this is ready to merge?

khaledhosny:
All good, thanks!

PR #249, opened 2022-09-26
[variations] Synthesize an empty 'gvar' if missing.

jfkthame:
If there is an 'fvar' table that defines variation axes, some rasterizers (e.g. Core Text) assume this means 'gvar' will be present, and fail if it is missing. To work around this, add an empty 'gvar' to such fonts.

This behavior is controlled by the --synthesize-gvar configuration option, so it can be disabled if not required in a given build.

(See discussion in https://github.com/simoncozens/parallax/issues/1 and https://bugzilla.mozilla.org/show_bug.cgi?id=1792214 for the background to this.)


src/ots.cc
```diff
@@ -737,6 +737,18 @@ bool ProcessGeneric(ots::FontFile *header,
...
   }
 
+#ifdef OTS_SYNTHESIZE_MISSING_GVAR
+  // If there was an fvar table but no gvar, synthesize an empty gvar to avoid
+  // issues with rasterizers (e.g. Core Text) that assume it must be present.
+  if (font->GetTable(OTS_TAG_FVAR) && !font->GetTable(OTS_TAG_GVAR)) {
+    ots::OpenTypeGVAR *gvar = new ots::OpenTypeGVAR(font, OTS_TAG_GVAR);
+    if (gvar->InitEmpty()) {
+      table_map[OTS_TAG_GVAR] = { OTS_TAG_GVAR, 0, 0, 0, 0 };
+      font->AddTable(gvar);
```
khaledhosny, line 747: We don’t check the return value of AddTable here (and doing so would be superfluous anyway), so is there is a point in AddTable checking for the existence of the table since it is only used here? May be it would be useful if we use use it more in the future, but I’m inclined to keep things simple and drop the check and return value, and worry about existing tables when we have to.

khaledhosny (approved):

src/ots.cc
jfkthame: Yeah, fair enough.... I suggest we instead `assert()` within AddTable that we're not adding a duplicate.

PR #252, opened 2022-11-17
[review] Set the correct values in the max size of glyf instructions warning.

bramstein:
I have a font giving an ots warning:

```
WARNING: glyf: Bytecode length is bigger than maxp.maxSizeOfInstructions 4937: 4937
```

This confused me a little until I looked at the code. I believe the warning should be printed before the `maxp` value is updated.


khaledhosny:
Thanks!

PR #256, opened 2023-03-30
Remove checks for Chain and Context rules that fail fonts for having glyph sequences that exceed the number of glyphs in the font.

garretrieger:
There's no reason that a glyph sequence can't exceed the number of glyphs in a font. These are sequences of glyphs and as such allow repetition of glyph ids. The current checks can cause small subsets of fonts to incorrectly fail OTS sanitization.

### DIFF
--- a/src/layout.cc
+++ b/src/layout.cc
@@ -446,7 +446,7 @@ bool ParseRuleSubtable(const ots::Font *font,
     return OTS_FAILURE_MSG("Failed to read rule subtable header");
   }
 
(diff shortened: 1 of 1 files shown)


behdad:
LGTM.

anthrotype (approved):

khaledhosny:
Thanks. See also #206. We probably should check all uses of number of glyphs in the font for any remaining wrong comparisons like this.

PR #267, opened 2023-11-10
[maxp] Attempt to improve fixup of bad 'maxp' version numbers.

jfkthame:
If the version number is less than 0x00010000 (e.g. an example reported at https://github.com/mozilla/pdf.js/issues/16839 apparently has 0x0100), it will be treated as version 0.5 and only the num_glyphs field is kept.

However, this can prevent the font working on Windows, which apparently requires the full version 1.0 maxp table for truetype fonts.

So if the version number was bad, but there is in fact enough data to parse as version 1.0, let's try correcting it to 1.0 rather than 0.5.

(Test not included because the example in the issue appears to be a derivative of a proprietary font.)

### DIFF
--- a/src/maxp.cc
+++ b/src/maxp.cc
@@ -29,39 +29,49 @@ bool OpenTypeMAXP::Parse(const uint8_t *data, size_t length) {
     return Error("numGlyphs is 0");
   }
 
-  if (version >> 16 == 1) {
-    this->version_1 = true;
-    if (!table.ReadU16(&this->max_points) ||
-        !table.ReadU16(&this->max_contours) ||
-        !table.ReadU16(&this->max_c_points) ||
-        !table.ReadU16(&this->max_c_contours) ||
-        !table.ReadU16(&this->max_zones) ||
-        !table.ReadU16(&this->max_t_points) ||
-        !table.ReadU16(&this->max_storage) ||
-        !table.ReadU16(&this->max_fdefs) ||
-        !table.ReadU16(&this->max_idefs) ||
-        !table.ReadU16(&this->max_stack) ||
-        !table.ReadU16(&this->max_size_glyf_instructions) ||
-        !table.ReadU16(&this->max_c_components) ||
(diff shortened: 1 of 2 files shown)


src/maxp.cc
```diff
@@ -29,6 +29,23 @@ bool OpenTypeMAXP::Parse(const uint8_t *data, size_t length) {
...
+    return true;
+  }
+
+  if (version != 0x00010000) {
+    Warning("Unrecognized version %08x, attempting to fix", version);
+    // If there's at least enough data to read as a version 1.0 table,
+    // we'll try that.
+    if (length >= 32) {
+      version = 0x00010000;
+    }
```
khaledhosny, line 46: I think this check is unnecessary. The table reads all do bounding checks. I’d simplify the code from here below to try to read the data and sets `version_1` based on whether this was successful or not.

src/ots.cc
```diff
@@ -762,12 +762,18 @@ bool ProcessGeneric(ots::FontFile *header,
...
 
   if (glyf && loca) {
     if (font->version != 0x000010000) {
       OTS_WARNING_MSG_HDR("wrong sfntVersion for glyph data");
       font->version = 0x000010000;
     }
+    if (!maxp->version_1) {
+      return OTS_FAILURE_MSG_TAG("wrong maxp version for glyph data",
+                                 OTS_TAG_MAXP);
+    }
```
khaledhosny, line 776: So if there is no `loca`/`glyf`, and there is no `cff` or `cff2`, we accept whatever version we have? Would a bitmap-only color font need a particular version? (or may be since these do not work on Windows any way, it does not matter).

jfkthame: It seems unlikely anything except `num_glyphs` would be relevant for such a font, so either version should be adequate. Whether any engine checks the version number is another matter... if so, we might want to set it accordingly, but for now I'd be inclined to just leave it untouched.

khaledhosny: Makes sense.

PR #269, opened 2023-11-24
Parsing Cff integer operands as signed integers

nagyatosz:
Cff specification defines the operands as signed (not unsigned) integers. This commit allows to keep the right signed value after corret parsing of negative integer values.

PR #270, opened 2023-12-13
[cff] correctly parse negative operands, but don't allow -ve offset

jfkthame:
This is @nagyatosz's patch from https://github.com/khaledhosny/ots/pull/269, plus a change to explicitly reject negative offsets. Although the spec does not seem to prohibit them, support in implementations is patchy, and AFAIK no mainstream tool or workflow will generate such a font. So it seems safer to reject them than to allow such an untested niche feature that implementations may mishandle.

@khaledhosny WDYT of doing something like this? We could reconsider if there's evidence of a real-world need to support this, though discussion indicated that not all engines would actually work with such a font.

### DIFF
--- a/src/cff.cc
+++ b/src/cff.cc
@@ -476,10 +476,17 @@ bool ParsePrivateDictData(
         if (operands.back().second != DICT_OPERAND_INTEGER) {
           return OTS_FAILURE();
         }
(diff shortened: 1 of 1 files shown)


khaledhosny (approved):
If we can add some test fonts that would be nice, otherwise LGTM.

PR #271, opened 2023-12-27
Fix a size_t/off_t mismatch build warning

kaidokert:
Fixes a build warning on builds where off_t and size_t are different.


khaledhosny:
Thanks!

Issue #200, opened 2019-10-08
Support bitmap fonts (EBDT, EBLC)


o5k:
In deciding to not support bitmap fonts at all, OTS completely ignores many cultures and the ability to write crisp small print:
![Japanese print in Bitmap and vector](https://kagari.moe/:/2019-10-08_14-55-04.png)
Furthermore, it may be needed for designers to use crisp rendering for stylish purposes as well.
As a major part of our web experience, OTS should look to support all languages and styles evenly!

khaledhosny:
I won’t have time to work on this, but patches are welcome. Browsers can also ask OTS tp pass-though these tables and this wouldn’t need any changes to OTS, but someone will have to ask browsers to do it.

mmulet:
I can patch this in, if it still needs doing.

khaledhosny:
Contributions are always welcome.

>

mmulet:
I made a Code Relay task for this https://github.com/code-relay-io/ots/blob/main/README.md . It will get done.

ra-om:
In on this. Don't want to be stuck to Firefox forever.

mmulet:
@ra-om We're on the ebdt branch right now. https://github.com/code-relay-io/ots/tree/ebdt You can jump right in.

ra-om:
In on this. Don't want to be stuck to Firefox forever.

flarn2006:
@mmulet I was wondering whether you found this issue the same way I found it. Then I looked back at where I found it and saw you created it. :P

Pomax:
@mmulet how's the work going? It would be great if OTS finally supported this =)

mmulet:
Code Relay has been on pause for a year and half (although we've been been attempting to restart it since January). 
Agreed that this one should really get done, so I think I will have to do it myself. I looked at the problem a few years ago and I believe that with copilot/gpt-4 these days, it shouldn't take too long. Depending on how much free time I have, I think finishing during this summer would be realistic.
(Side note: Are you the pomax of the bezier primer? If so, I just want to say that the primer is a goldmine of information and you have my heartfelt thanks for making it.)

Pomax:
I am, always good to know it's been useful.

mmulet:
Here is an update because it's been about 6 months and past my summer estimate. 

This issue has been on my todo list, but life and work keep on getting in the way. This issue keeps on getting pushed back further and further on my list. At this rate, I estimate will be finishing by the end of January.

Sorry for the delay!

[...]

PR #273, opened 2024-01-31
add EBDT and EBLC table support.

mmulet:
This pull request adds basic parsing and validation for EBDT and EBLC tables in order to support bitmap fonts issue #200 .

It's been tested against the batang.ttc and guilm.ttc fonts supplied with the Korean Supplemental fonts in windows (after some small modifications to get the batang.ttc and guilm.ttc fonts to pass GSUB and CMAP sanitization). Those fonts 
used Image formats 2 and 7 and EBLC index subtables 3 and 1. I'm still looking for other fonts that utilize the other index and image formats for testing.

Let me know what else has to be done, to support bitmap fonts.


dimdenGD:
Could someone review/merge this? It'd be great to add support for bitmap fonts.

mmontag:
@khaledhosny friendly ping

khaledhosny:
It is a quite some code and I don’t have much time to review it currently. But a first step would to provide a set of test fonts that we can add to the test suite, otherwise it is a big chunk of untested code.

PR #279, opened 2024-03-25
Fix potential arithmetic overflows

jfkthame:
This aims to fix some potential arithmetic overflows, originally reported to Firefox and Chrome.


khaledhosny:
Thanks!

PR #229, opened 2020-12-09
Increase limits for decompressed tables and files

josh-hadley:
- add `OTS_MAX_DECOMPRESSED_FILE_SIZE` and `OTS_MAX_DECOMPRESSED_TABLE_SIZE` constants
- replace previous limit values with those constants
- minor updates to the related failure messaging

Closes #219

Issue #219, opened 2020-10-21
What is the rationale for limiting WOFF/WOFF2 uncompressed table length to 30MB?


josh-hadley:
Per
https://github.com/khaledhosny/ots/blob/1fe1227f2c251c2656106a004bbed83b41936b63/src/ots.cc#L633-L635

uncompressed (WOFF, WOFF2) tables are limited to 30MB. Table length limit otherwise is 1GB. What is the rationale for the 30MB limit for compressed -> uncompressed?

khaledhosny:
I have no idea. The WOFF check is from the first implementation in a6a1b6de which predates my fork. The WOFF2 check is even worse as the whole font is checked against the 30 MB limit, from the first implementation in 96d74fe3 which also predates my fork.

You would need someone from Chrome to track down that decision.

anthrotype:
Hm, the arbitrary limit looks strange indeed. I was told there are some large CJK fonts among macOS system fonts (e.g. `HanziPenTC-W3`) which fail OTS when compressed as WOFF or WOFF2. But they _pass_ if sanitized as uncompressed TTCs.

josh-hadley:
@anthrotype that's exactly the situation I'm facing (large CJK fonts). No issues with the raw/uncompressed font, but the WOFF & WOFF2 files fail because of these limits.

behdad:
The limits were put in place supposedly to protect against gzip bombs in WOFF as well as classic myopic "who needs webfonts with tables > 30MB anyway"...

drott:
The review of a6a1b6debe569cd69a02fc5089ac8e69f4d9ba5f was here: https://codereview.appspot.com/946044 - upgrading this limit during the review from 16MB to 30MB. - In the review, Yusuke Sato compared the initial limit of 16MB to a set of Windows fonts:

> ```
> Some of unmalicious fonts seem to exceed this limitation. How about using 30MB
> or something?
> 
> fonts$ find . -size +16M -name '*tf' -ls
> -rwxr-xr-x   1 yusukes        27382464 Aug 25  2009 ./windows/mingliu0.ttf
> -rwxr-xr-x   1 yusukes        27382472 Aug 25  2009 ./windows/mingliu1.ttf
> -rwxr-xr-x   1 yusukes        27396928 Aug 25  2009 ./windows/mingliu2.ttf
> -rwxr-xr-x   1 yusukes        23278008 Jul  1  2009 ./mac/Arial\ Unicode.ttf
> ```

[...]

PR #284, opened 2024-09-26
Set correct max_size for woff2 decompress output buffer

Jackyjjc:
This pull request addresses the issue in https://github.com/khaledhosny/ots/issues/219.

The [original PR](https://github.com/khaledhosny/ots/pull/229) did not fully fix the issue of 30MB limit.

The current code still fails for fonts >30MB uncompressed when calling [woff2::ConvertWOFF2ToTTF](https://github.com/khaledhosny/ots/blob/27d63ae4d8988619e0a661bf40551c47da47f3fa/src/ots.cc#L547).

The issue is that the constructor of WOFF2StringOut [sets the maxSize](https://github.com/google/woff2/blob/0f4d304faa1c62994536dc73510305c7357da8d4/src/woff2_out.cc#L14) of the buffer to `kDefaultMaxSize`, which is also set to [30MB](https://github.com/google/woff2/blob/0f4d304faa1c62994536dc73510305c7357da8d4/include/woff2/output.h#L20)

To properly fix this, we will need to ensure the maxSize of the output buffer is set correctly.

### DIFF
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -544,6 +544,7 @@ bool ProcessWOFF2(ots::FontFile *header,
 
   std::string buf(decompressed_size, 0);
   woff2::WOFF2StringOut out(&buf);
+  out.SetMaxSize(decompressed_size);
   if (!woff2::ConvertWOFF2ToTTF(data, length, &out)) {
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks!

PR #288, opened 2024-10-04
[glyf] Respect WE_HAVE_INSTRUCTIONS on any component of a composite.

jfkthame:
Noticed in passing that the `OpenTypeGLYF::ParseCompositeGlyph` method doesn't quite follow [the spec](https://learn.microsoft.com/en-gb/typography/opentype/spec/glyf#composite-glyph-description), which tells us that

> If the flag [i.e. WE_HAVE_INSTRUCTIONS] is set on any component glyph, then a uint16 value is read immediately after the last component glyph to get the byte length for instructions.

so it's not sufficient to just check the `flags` that came from the last component.

### DIFF
--- a/src/glyf.cc
+++ b/src/glyf.cc
@@ -262,6 +262,7 @@ bool OpenTypeGLYF::ParseCompositeGlyph(
     ComponentPointCount* component_point_count) {
   uint16_t flags = 0;
   uint16_t gid = 0;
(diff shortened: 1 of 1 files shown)


jfkthame:
This is included as part of the most recent update to https://github.com/khaledhosny/ots/pull/289, so we can close this as a standalone PR.

PR #289, opened 2024-10-04
[glyf] Attempt to drop empty components from composite glyphs.

jfkthame:
This is because such components cause DirectWrite to fail to render the composite at all, even though an empty component should (presumably) have no visible effect.

Reported in the context of pdf.js at https://github.com/mozilla/pdf.js/issues/18848, but experimentation indicates that Windows/DWrite fails to paint such glyphs in other contexts as well -- e.g. when using Character Map to view an installed font.

(I don't think there's any real use case for such components, and their presence is probably an anomaly/artifact of a particular font creation/subsetting/embedding workflow. Still, if OTS can fix them up, this will best serve end users.)

The current patch discards empty components from composite glyphs *unless* the empty glyph is the last component. In that case, we can't just skip it, because the preceding component (which will now be last) has the MORE_COMPONENTS flag, and that's no longer true. I have not seen cases in the wild where this issue affects the final component, but it's presumably possible (and testing indicates that DirectWrite would fail in that case too), so this would be a good followup enhancement. Currently the code just issues a warning if such a case occurs.


tvsinh:
if OTS can fix them up, this will best serve end users

jfkthame:
With the second commit here, the limitation mentioned in https://github.com/khaledhosny/ots/pull/289#issue-2566824198 (unable to drop the last component of the composite) is no longer true; we handle this case, fixing up component flags as needed.

tvsinh:
This could be a useful patch for many end users.

khaledhosny:
Thanks!

Issue #290, opened 2024-10-25
[`CFF2`] Lift `uint16 VariationStore.length` limitation


behdad:
See this for details:
https://github.com/harfbuzz/boring-expansion-spec/issues/159

I suggest we implement that already. That is, if the `VariationStore` struct's length field is 65,535, then allow the embedded `ItemVariationStore` to reach out to the rest of the font file.

@justvanrossum has been building a font that hit that limit.

We are going to also implement this change in FontTools. HarfBuzz already renders such fonts fine, and I will test FreeType and report.

khaledhosny:
What does Microsoft and Apple rasterizer do with such fonts? We wouldn’t usually allow fonts they can’t handle, and specially if they don’t fail gracefully. cc @jfkthame

jfkthame:
> What does Microsoft and Apple rasterizer do with such fonts? We wouldn’t usually allow fonts they can’t handle, and specially if they don’t fail gracefully. cc @jfkthame

Yeah, that's the same question I was going to ask. We don't want OTS accepting fonts that will then cause issues in either DirectWrite or Core Text.

Are there some examples of such fonts that folks could test with the various rasterizers?

behdad:
> > What does Microsoft and Apple rasterizer do with such fonts? We wouldn’t usually allow fonts they can’t handle, and specially if they don’t fail gracefully. cc @jfkthame
> 
> Yeah, that's the same question I was going to ask. We don't want OTS accepting fonts that will then cause issues in either DirectWrite or Core Text.

I can install the font in on the Mac just fine, and renders fine. I cannot test on Windows.

> Are there some examples of such fonts that folks could test with the various rasterizers?

I have a test font from @justvanrossum that is proprietary. I can get it sorted to share with you or @khaledhosny if you can test on Windows.

jfkthame:
Windows 10 refuses to install the font; it tells me it's "not a valid font file".

I was able to load it as a webfont resource in Firefox after hacking the OTS code to skip validation of the CFF2 table, and then it does seem to render the glyphs as expected (including weight variations), though I don't know how to test whether it would be able to use the whole of the VariationStore, or only whatever's in the first 64K.

(Just patching ParseVariationStore wasn't sufficient to get through OTS, as the CFF2 table was also rejected for invalid CharStrings; I had to make it skip CFF2 validation altogether.)

behdad:
[...]

PR #291, opened 2024-10-25
[CFF2] Lift uint16 VariationStore.length limitation

behdad:
Fixes https://github.com/khaledhosny/ots/issues/290


src/cff.cc
```diff
@@ -588,19 +588,23 @@ bool ParsePrivateDictData(
...
 
+  unsigned length = encoded_length;
+
   // Empty VariationStore is allowed.
   if (!length) {
     return true;
   }
 
-  if (length > table.remaining()) {
+  if (length != 65536 && length > table.remaining()) {
```
jfkthame, line 604: I think you meant `length != 65535 ...` here.

behdad: Right. Thanks for catching!

behdad:
The logic was actually wrong. PTAL.

PR #292, opened 2025-01-02
Show invalid table tags as hex in error messages.

jfkthame:
This avoids the risk of the returned message including NUL or other non-printing characters.

(See https://bugzilla.mozilla.org/show_bug.cgi?id=1937227 for the example that motivated this change.)


khaledhosny:
Thanks!

PR #297, opened 2025-06-18
Introduce new TABLE_ACTION_SANITIZE_SOFT action.

jfkthame:
This action (not used as the default for any table tags) performs the same table-specific Parse() as TABLE_ACTION_SANITIZE, but if parsing fails, it will drop the table but _not_ return a failure status to the overall processing.

This can be used by a client that wishes to continue processing and using a resource even if certain tables may fail sanitization and be discarded.

It differs from SANITIZE in that failure to parse does not completely stop processing of the resource and return an overall failure.

It differs from PASS_THRU in that the table is subject to sanitization, reporting any errors found, and will be dropped if it fails, rather than silently accepted.

It differs from DROP in that the table is only dropped if (unrecoverable) errors are found by Parse(); if the table is valid, it will be retained.

### DIFF
--- a/include/opentype-sanitiser.h
+++ b/include/opentype-sanitiser.h
@@ -172,7 +172,9 @@ enum TableAction {
   TABLE_ACTION_DEFAULT,  // Use OTS's default action for that table
   TABLE_ACTION_SANITIZE, // Sanitize the table, potentially dropping it
   TABLE_ACTION_PASSTHRU, // Serialize the table unchanged
-  TABLE_ACTION_DROP      // Drop the table
+  TABLE_ACTION_DROP,     // Drop the table
+  TABLE_ACTION_SANITIZE_SOFT, // Sanitize the table, but without failing overall
+                              // sanitzation even if this table fails/is dropped
 };
 
 class OTSContext {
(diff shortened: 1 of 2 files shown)


jfkthame:
The motivation for this is that we'd like to try using it in Firefox for OpenType Layout tables, where we currently have to choose between PASS_THRU (which silently ignores invalid tables) or SANITIZE (which causes the entire resource to be rejected if errors are found). The new option will allow us to detect and report errors, while still allowing the (remainder of the) resource to load.

This could result in broken rendering in complex-script cases where OTL processing is necessary, but in practice we see cases where (for example) a font has been incorrectly subsetted so as to just use a few symbols; associated OTL tables were not correctly adjusted, but are not actually relevant to the handful of glyphs being used. So dropping the invalid OTL tables but loading the residual font is a more user-friendly outcome than rejecting the resource altogether.

khaledhosny:
An alternative approach would be returning to the (very) old behavior of not considering layout tables required. IIRC Chrome currently passes thru layout tables, so it is only Firefox that sanitizes them (I believe only in non-release builds), so if Firefox wants to make them non-required, then there is no much point in keeping the current behavior.

jfkthame:
Ah, right - yes, that was basically equivalent. I think I prefer this, however, because it lets the client decide which behavior to apply, rather than hardcoding it in OTS.

behdad:
I have concerns that we'd again start getting bug reports against shaping, where the font has been modified by ots by dropping the layout tables.

PR #305, opened 2026-03-27
[colr] Avoid double-reading the base glyph list entries.

jfkthame:
There's no need to save and reset the subtable offset, and re-read the base glyph list entries, as we have saved them in state.baseGlyphMap and can just iterate over that to validate the records.


khaledhosny:
Thanks!

PR #306, opened 2026-03-27
[cff] Use CheckOffset to validate private_offset.

jfkthame:
The current code checks for offset > length but fails to check for a negative value.

### DIFF
--- a/src/cff.cc
+++ b/src/cff.cc
@@ -556,7 +556,7 @@ bool ParsePrivateDictData(
           return OTS_FAILURE();
         }
         uint16_t k = out_cff->region_index_count.at(vsindex);
-          
+
         if (operands.back().first > static_cast<uint16_t>(0xffff) || operands.back().first < 0){
           return OTS_FAILURE();
         }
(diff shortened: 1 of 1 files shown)


src/cff.cc
khaledhosny, line 402: The fix is fine, but I think the root cause is this pattern here. `ots::Buffer` reads are all bounds-checked, but here we are creating a new buffer without any bounds-checking. I think all uses of `ots::Buffer::buffer()` like this are dangerous. We should not allow this at all, and instead make `ots::Buffer::buffer()` take an offset and length, bounds-check them and return an `ots:Buffer` (maybe even name it `sub_buffer()` or something similar.

jfkthame: Yes, that sounds sensible. It'll clearly be a substantially bigger/more invasive patch, though. So unless you're planning on doing it right away (sorry, I don't think I have time to tackle it right now), I'd suggest we merge this as a short-term fix for the immediate issue.

jfkthame: (FTR, we've applied this downstream in Firefox already, but obviously happy to see it replaced with a more systemic fix in a future version.)

khaledhosny:
Lets merge then, and keep the bigger change for later.

PR #307, opened 2026-04-07
Fix format string

andyleiserson:


khaledhosny:
What issue are you having here that needs this fix?

andyleiserson:
It is not causing issues, just code cleanup.

khaledhosny:
Thanks!

PR #309, opened 2026-04-12
Fix null pointer dereference in ProcessGeneric() with TABLE_ACTION_PASSTHRU

### LINKED ISSUE #308: [Fuzzing] Null Pointer Dereference: maxp dereference in ProcessGeneric() with TABLE_ACTION_PASSTHRU
## Severity

**High** — denial of service. Any application using OTS with `TABLE_ACTION_PASSTHRU` can be crashed by a crafted font file.

## Affected Function / File

- **File:** `src/ots.cc`
- **Function:** `ProcessGeneric()` at lines 781 and 798
- **Entry points:** `ProcessTTF()`, `ProcessWOFF()`

## Root Cause

`ProcessGeneric()` retrieves the `maxp` table via `GetTypedTable(OTS_TAG_MAXP)` at line 773-774. In `TABLE_ACTION_PASSTHRU` mode, tables are stored as `TablePassthru` objects whose `Type()` returns a passthrough tag rather than `OTS_TAG_MAXP`. This causes `GetTypedTable()` to return `NULL` even when a `maxp` table is present in the font. The code then dereferences the null pointer at two locations:

- **Line 781:** `if (!maxp->version_1)` — reached when font has glyf+loca tables
- **Line 798:** `if (maxp->version_1)` — reached when font has CFF/CFF2 tables

The crash is a deterministic SEGV at address `0x1c` (offset of `version_1` within the `OpenTypeMAXP` struct), confirming a null `this` pointer.

**Introduced in:** Commit `83ff34d` (2023-11-10), merged via PR #267 (commit `be94458`, 2023-11-12).

## Crash Signature

```
ERROR: AddressSanitizer: SEGV on unknown address 0x00000000001c
The signal is caused by a READ memory access.
Hint: address points to the zero page.
```

### Representative Stack Trace

```
#0 ProcessGeneric()      src/ots.cc:781:9
#1 ProcessTTF()          src/ots.cc:282:10
#2 OTSContext::Process()  src/ots.cc:1203:14
```

## Reproduction Steps

1. Build a fuzzer harness that calls `OTSContext::Process()` with all table actions set to `TABLE_ACTION_PASSTHRU`:

```cpp
#include <cstdint>
#include <cstdlib>
#include "opentype-sanitiser.h"
#include "ots-memory-stream.h"

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
  ots::OTSContext context;
  for (int tag = 0; tag < 4; tag++) {
    for (int i = 0; i < 256; i++) {
      context.SetTableAction(OTS_TAG(
        ('a' + ((tag >> 0) & 0xf)),
        ('a' + ((tag >> 4) & 0xf)),
        ('a' + ((i >> 0) & 0xf)),
        ('a' + ((i >> 4) & 0xf))
      ), ots::TABLE_ACTION_PASSTHRU);
    }
  }
  // Set all tables to passthrough
  context.SetTableAction(0, ots::TABLE_ACTION_PASSTHRU);

  ots::ExpandingMemoryStream output(size, size * 2);
  context.Process(&output, data, size);
  return 0;
}
```

2. Compile with sanitizers:
```bash
clang++ -g -O1 -fsanitize=fuzzer,address \
  -I include -I src -I build -DHAVE_CONFIG_H \
  fuzz_ots_passthru.cc build/libots.a \
  build/subprojects/woff2-1.0.2/libwoff2_decoder.a \
  build/subprojects/woff2-1.0.2/libwoff2_common.a \
  -lz -lbrotlidec -lbrotlicommon -llz4 \
  -o fuzz_ots_passthru
```

3. Run the fuzzer — crashes within seconds, or provide any minimal TTF/WOFF font with glyf+loca or CFF tables and a missing/passthrough maxp.

## Exploitability

**DoS only** — not exploitable for code execution. The crash is a deterministic read from a fixed near-null address (`0x1c`), which is always unmapped. No heap corruption, no controlled write, no info leak.

## Suggested Fix

Add a null check for `maxp` after the `GetTypedTable()` call at line 774, before either dereference site:

```cpp
ots::OpenTypeMAXP *maxp = static_cast<ots::OpenTypeMAXP*>(
  font->GetTypedTable(OTS_TAG_MAXP));

if (!maxp) {
  return OTS_FAILURE_MSG_TAG("missing maxp table", OTS_TAG_MAXP);
}
```

This is consistent with how other required tables are validated elsewhere in the codebase.

## Notes

- The default sanitize mode (`TABLE_ACTION_DEFAULT`) is **not** affected because it returns properly typed table objects from `GetTypedTable()`.
- The existing `ots-fuzzer` (which uses the default mode) never found this bug because it doesn't exercise the passthrough code path.
- Found via automated fuzzing with libFuzzer + AddressSanitizer on commit `8510bfd`.

parasol-aser:
## Summary

- **Fix null pointer dereference in `ProcessGeneric()` (`src/ots.cc`)**: Add null checks for `maxp` after `GetTypedTable(OTS_TAG_MAXP)` in both the `glyf/loca` and `cff/cff2` branches. In passthrough mode, `TablePassthru` uses type `0` instead of the actual tag, so `GetTypedTable()` returns `NULL`. The pointer was dereferenced without a null check, causing a deterministic SEGV (DoS crash).
- **Fix same bug pattern in `OpenTypeGLYF::Parse()` (`src/glyf.cc`)**: Add null check for `name` after `GetTypedTable(OTS_TAG_NAME)` before calling `name->IsTrickyFont()`. Crashes when `name` table is passthrough while `glyf` is sanitized.
- **Add comprehensive regression tests** (`tests/passthru_test.cc`): 17 test cases covering all-passthrough, selective passthrough (maxp-only, name-only, combined), drop actions, boundary inputs (empty, truncated, garbage), and sequential context switching.

Fixes #308

## Test plan

- [ ] `passthru_test` — new test suite passes (all 17 cases: no SEGV, graceful failure where expected)
- [ ] Existing `meson test` suite passes with no regressions (ots-fuzzer, ots-sanitize on bad fonts, ots-idempotent on good fonts)
- [ ] Bitmap-only fonts in passthrough mode are NOT rejected (null check is per-branch, not early-return)
- [ ] Build with ASan to verify no memory errors in passthrough paths

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/glyf.cc
+++ b/src/glyf.cc
@@ -462,15 +462,12 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
       GetFont()->GetTypedTable(OTS_TAG_LOCA));
   OpenTypeHEAD *head = static_cast<OpenTypeHEAD*>(
       GetFont()->GetTypedTable(OTS_TAG_HEAD));
-  if (!maxp || !loca || !head) {
-    return Error("Missing maxp or loca or head table needed by glyf table");
-  }
-
   OpenTypeNAME *name = static_cast<OpenTypeNAME*>(
       GetFont()->GetTypedTable(OTS_TAG_NAME));
-  if (!name) {
-    return Error("Missing name table needed by glyf table");
+  if (!maxp || !loca || !head || !name) {
+    return Error("Missing maxp or loca or head or name table needed by glyf table");
   }
+
   bool is_tricky = name->IsTrickyFont();
 
   this->loca = loca;
--- a/src/glyf.cc
+++ b/src/glyf.cc
@@ -468,6 +468,9 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
 
   OpenTypeNAME *name = static_cast<OpenTypeNAME*>(
       GetFont()->GetTypedTable(OTS_TAG_NAME));
+  if (!name) {
+    return Error("Missing name table needed by glyf table");
+  }
   bool is_tricky = name->IsTrickyFont();
 
   this->loca = loca;
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -774,6 +774,9 @@ bool ProcessGeneric(ots::FontFile *header,
     font->GetTypedTable(OTS_TAG_MAXP));
 
   if (glyf && loca) {
+    if (!maxp) {
+      return OTS_FAILURE_MSG_TAG("missing maxp table", OTS_TAG_MAXP);
+    }
     if (font->version != 0x000010000) {
       OTS_WARNING_MSG_HDR("wrong sfntVersion for glyph data");
       font->version = 0x000010000;
@@ -787,6 +790,9 @@ bool ProcessGeneric(ots::FontFile *header,
     if (cff2)
        cff2->Drop("font contains both CFF and glyf/loca tables");
   } else if (cff || cff2) {
+    if (!maxp) {
+      return OTS_FAILURE_MSG_TAG("missing maxp table", OTS_TAG_MAXP);
+    }
     if (font->version != OTS_TAG('O','T','T','O')) {
       OTS_WARNING_MSG_HDR("wrong sfntVersion for glyph data");
       font->version = OTS_TAG('O','T','T','O');
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -770,10 +770,10 @@ bool ProcessGeneric(ots::FontFile *header,
   ots::Table *loca = font->GetTable(OTS_TAG_LOCA);
   ots::Table *cff  = font->GetTable(OTS_TAG_CFF);
   ots::Table *cff2 = font->GetTable(OTS_TAG_CFF2);
-  ots::OpenTypeMAXP *maxp = static_cast<ots::OpenTypeMAXP*>(
-    font->GetTypedTable(OTS_TAG_MAXP));
 
   if (glyf && loca) {
+    ots::OpenTypeMAXP *maxp = static_cast<ots::OpenTypeMAXP*>(
+      font->GetTypedTable(OTS_TAG_MAXP));
     if (!maxp) {
       return OTS_FAILURE_MSG_TAG("missing maxp table", OTS_TAG_MAXP);
     }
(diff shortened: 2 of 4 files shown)


src/ots.cc
```diff
@@ -774,6 +774,9 @@ bool ProcessGeneric(ots::FontFile *header,
     font->GetTypedTable(OTS_TAG_MAXP));
 
   if (glyf && loca) {
+    if (!maxp) {
```
khaledhosny, line 777: Move getting the `maxp` table here so that no future code would try accessing it without NULL check.

src/ots.cc
```diff
@@ -787,6 +790,9 @@ bool ProcessGeneric(ots::FontFile *header,
     if (cff2)
        cff2->Drop("font contains both CFF and glyf/loca tables");
   } else if (cff || cff2) {
+    if (!maxp) {
```
khaledhosny, line 795: Same here.

khaledhosny (changes requested):

src/glyf.cc
```diff
@@ -468,6 +468,9 @@ bool OpenTypeGLYF::Parse(const uint8_t *data, size_t length) {
 
   OpenTypeNAME *name = static_cast<OpenTypeNAME*>(
       GetFont()->GetTypedTable(OTS_TAG_NAME));
```
khaledhosny, line 466: Move this code up, and check for name table with the other table checks above.

parasol-aser:
Thanks for the review! Addressed in 201abc2:

- **`src/ots.cc`**: Moved the `maxp = GetTypedTable(OTS_TAG_MAXP)` call inside each branch (`if (glyf && loca)` and `else if (cff || cff2)`), so `maxp` is no longer in scope for the bitmap-only or error branches. Any future code added to the outer scope can't dereference it without a null check.
- **`src/glyf.cc`**: Moved the `name` table lookup up alongside `maxp`/`loca`/`head` and folded it into the single combined null-check, matching the existing pattern.

All existing tests and the new `passthru_test` still pass.

khaledhosny (approved):

khaledhosny:
Thanks!

PR #310, opened 2026-04-23
[cff] Validate ordering of hint operators in charstring.

jfkthame:
The spec requires that all stems are defined at the beginning of the charstring (before any contours), and that the hint ops are used in a specific order (HSTEMs before VSTEMs; CNTRMASK before HINTMASK).

https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf#G3.31003

NB: We don't currently enforce the spec requirement that if HINTMASK is present in the charstring, the stems must be defined using the ...HM variants of HSTEM and VSTEM operators.


khaledhosny:
Thanks!

PR #314, opened 2026-04-27
[cff] Update Operand handling.

### LINKED ISSUE #312: Possible CFF1 validation bypass when encountering BCD operands
Spun-off from: https://crbug.com/505807063

We received a report from an AI based vulnerability search tool in Chromium which complains about BCD operands being inserted as dummies, which contradicts the purpose of validation in `ParseDictData`. Report below.  Please take a look - there might be an option to improve CFF validation. 

### Summary
A vulnerability exists in the OpenType Sanitizer (OTS) CFF1 parser where Binary Coded Decimal (BCD) operands in the Top DICT can be used to bypass structural validation for `charset` and `Encoding` tables. This bypass occurs because OTS uses a dummy value of `0` for BCD operands and fails to verify the operand type before checking for predefined constants.

### Root Cause Analysis
In `third_party/ots/src/src/cff.cc`, the `ParseDictDataBcd` function processes BCD operands. Since OTS does not currently compute the actual numerical value of BCD numbers, it stores all such operands using a dummy integer value of `0` with the type `DICT_OPERAND_REAL`:

```cpp
// third_party/ots/src/src/cff.cc
operands.push_back(std::make_pair(static_cast<uint32_t>(0),
                                   DICT_OPERAND_REAL));
```

When parsing the Top DICT, operators `15` (`charset`) and `16` (`Encoding`) check if the provided operand matches a predefined value (0, 1, or 2 for charset; 0 or 1 for Encoding) representing a built-in table. However, this check is performed on the value before verifying the operand type:

```cpp
// third_party/ots/src/src/cff.cc (inside ParseDictData)
      case 15:  // charset
        if (operands.size() != 1) {
          return OTS_FAILURE();
        }
        if (operands.back().first <= 2) {
          // predefined charset, ISOAdobe, Expert or ExpertSubset, is used.
          break;
        }
        if (!CheckOffset(operands.back(), table.length())) {
          return OTS_FAILURE();
        }
```

Because the dummy value for a BCD operand is `0`, it satisfies the `first <= 2` check for `charset` (and `first <= 1` for `Encoding`), causing the parser to `break` early. This entirely skips the `CheckOffset` call, which normally ensures the operand is an integer and that the offset is within the table's bounds. 

Consequently, the `charset_offset` (or `encoding_offset`) variable remains initialized at `0`, and the structural validation loop for these tables is completely skipped.

### Impact
OTS serves as the primary security boundary against malicious fonts. By bypassing this validation step, a malformed font is passed directly to the platform's native font engine.

### Potential Reproduction Steps
*Note: The following reproduction steps are suggested/potential, as they were identified via static analysis tooling and have not yet been executed as a working proof of concept.* 

1. Construct an OTF font with a CFF1 table.
2. In the Top DICT, encode the `charset` (operator 15) or `Encoding` (operator 16) operand as a BCD real (starting with byte `0x1E`). Set the BCD value to an offset pointing to an intentionally malformed data structure within the font file.
3. Embed the malicious font in a webpage using a `@font-face` CSS rule.
4. Have a Chromium-based browser navigate to the page.
5. Observe that OTS successfully validates the font (due to the bypass) and forwards it to the OS font engine, even though the font is malformed.

### Suggested Fix
Update `third_party/ots/src/src/cff.cc` to explicitly enforce that operands for `charset` and `Encoding` are integers before checking their values against predefined constants.

For example, in `case 15:` and `case 16:`:
```cpp
if (operands.back().second != DICT_OPERAND_INTEGER) {
  return OTS_FAILURE();
}
if (operands.back().first <= 2) {
  break;
}
```

jfkthame:
This replaces the pair<int32_t, DICT_OPERAND_TYPE> with an Operand class that adds assertions when reading back the value, to ensure we don't store as one type and then read the value as another.

It also evaluates and stores decimal (BCD-encoded) values, rather than just discarding them, although we don't actually check their values anywhere at this point.

Also verifies that operands have INTEGER type in a couple of places where this is expected but wasn't being checked.

Fixes #312.

### DIFF
--- a/src/cff.cc
+++ b/src/cff.cc
@@ -5,6 +5,7 @@
 #include "cff.h"
 
 #include <cstring>
+#include <cmath>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks!

PR #318, opened 2026-05-30
[hhea/vhea] Stop forcing caretOffset to zero for non-italic fonts

### LINKED ISSUE #203: Unreasonable coupling of `macStyle.italic` vs `hhea.caretOffset` value
As discussed in https://github.com/adobe-type-tools/afdko/issues/1095, it seems [ots's check of `hhea.caretOffset`](https://github.com/khaledhosny/ots/blob/968e5f98ded4b90105ce78724e7af6940702b84f/src/metrics.cc#L52-L57) may be a bit overzealous.

miguelsousa:
The check in **metrics.cc** reset `caretOffset` to 0 whenever the `head.macStyle` italic bit was not set. Per the OpenType spec a zero `caretOffset` for non-slanted fonts is only a recommendation, not a requirement, and tools like makeOTF legitimately emit non-zero values. OTS was silently destroying valid font data.

Remove the coupling so `caretOffset` passes through untouched. The `head` table lookup (and its include) are no longer needed.

Fixes #203.

### DIFF
--- a/src/metrics.cc
+++ b/src/metrics.cc
@@ -4,7 +4,6 @@
 
 #include "metrics.h"
 
-#include "head.h"
 #include "maxp.h"
 
 // OpenType horizontal and vertical common header format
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks!

PR #325, opened 2026-06-15
[ots] Accept bitmap-only fonts with EBDT/EBLC

P-DOX:
**Description**
**Context :** This is part of work to support bitmap-only fonts (CBDT/CBLC color bitmaps and EBDT/EBLC monochrome bitmaps) in Chromium. Chromium currently rejects such fonts at the OTS pass before they reach the font stack, so this change is a prerequisite for the wider effort.
 
Extend the existing bitmap-only pass-through (CBDT/CBLC) to also accept fonts whose only glyph data is the monochrome embedded bitmap pair EBDT/EBLC. Previously such fonts were rejected with "no supported glyph data table(s) present" even when the caller asked OTS to pass bitmap-only fonts through.

### DIFF
--- a/src/ots.cc
+++ b/src/ots.cc
@@ -807,8 +807,10 @@ bool ProcessGeneric(ots::FontFile *header,
       OTS_WARNING_MSG_HDR("fixing incorrect maxp version for CFF font");
       maxp->version_1 = false;
     }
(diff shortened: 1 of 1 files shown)


drott (approved):
Non owner LGTM, let's hear it from @khaledhosny

khaledhosny:
Thanks!

PR #326, opened 2026-07-02
Add VARC (Variable Composites / Components) table support

behdad:
Adds a sanitizer for the `VARC` table from the [boring-expansion-spec](https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md), which encodes variable-composite glyphs combinable with `glyf`/`gvar` or `CFF2`.

Following the `CFF`/`COLR` pattern, the table is deeply validated in `Parse` and re-emitted verbatim in `Serialize` (no reconstruction).

The parser validates:
- the table header and its five sub-table offsets;
- the `MultiItemVariationStore` — sparse variation regions (axis coordinates checked as in the existing `ItemVariationStore` code), item-variation-data sub-tables and their delta-set (CFF2-style) indices, collecting per-outer delta-set counts;
- the `ConditionList` and all five `Condition` formats, recursively with a depth limit;
- the `axisIndicesList`, decoding each `TupleValues` entry to determine how many axis values each referring component must supply;
- every `VarCompositeGlyph` record — `uint32var` flags, 16/24-bit gid, condition index, axis indices/values, axis-values/transform variation indices, the transform fields, and reserved `uint32var` fields.

Cross-references are bounds-checked against `maxp`/`fvar` and the collected counts: component gid, condition index, `axisIndices` index (and the axis-value count derived from it), `VarIdx` (outer/inner), and region/axis indices. New helpers cover `uint32var`, dual-mode `TupleValues` (count-known and length-known), and the CFF2-style `Index` (32-bit count).

Adds three glyf-flavored good-font tests, one exercising all five `Condition` formats. The full test suite passes; round-trip output is byte-identical to the input, and deliberately corrupted VARC tables are rejected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


behdad:
@khaledhosny can you review this please?

behdad:
Hi, I’m Codex, an AI coding agent from OpenAI. I reviewed this PR for sanitizer blockers and correctness bugs.

I would block merging on the first four findings:

1. **Blocker — exponential condition-validation DoS.**
   `ConditionAnd`/`ConditionOr` recursively validate every referenced child, but only recursion depth is limited. Reusing the same child twice at each level produces exponential work. A 309-byte VARC with 29 levels took **3.32 seconds** to sanitize; slightly deeper inputs can effectively hang OTS. Add memoization or a global condition-edge/work budget.

2. **Blocker — variation delta payloads are not sanitized.**
   `ParseMultiItemVariationData` only validates the enclosing CFF2 Index; it never parses its `TupleValues` objects. I changed a 14-byte delta set to declare a 64-byte run: OTS reported success and preserved it, while fontTools rejected it as truncated. Each object must be decoded within its own bounds. Tuple cardinality should also be retained so VarIdx uses can be checked against the required axis/transform/value tuple length.

3. **Blocker — `axisIndicesList` values are never validated.**
   The parser counts tuple entries but discards their values. In a font with eight `fvar` axes, I changed `[2,3,4]` to `[127,3,4]`; OTS sanitized it successfully and retained axis 127. Decode these values and require every index to be below `fvar.axisCount`. Component axis values are similarly only skipped rather than semantically checked.

4. **Blocker — zero glyph records disables the required coverage-count check.**
   A count of zero is passed to `ParseCoverageTable`, where zero means “do not check.” Changing the fixture from four glyph records to zero while leaving four covered glyphs was accepted. Counts above `0xFFFF` also disable validation. Coverage indices select glyph records, so the counts must always agree.

5. **Bug — invalid five-byte `uint32var` values are truncated.**
   The implementation acknowledges that only `0xF0` is valid but accepts `0xF1`–`0xFF`, discarding the high bits. `F1 00 00 00 00`, representing `2^32`, was accepted as zero. Reject any five-byte lead other than `0xF0`; ideally also enforce canonical encodings.

6. **Bug — Condition format 1 accepts coordinates outside the normalized range.**
   The new parser checks only `axisIndex`. A condition spanning raw `[-0x8000, 0x7FFF]` was accepted, although these condition values must be between −1 and +1. The existing OTS condition validator already performs this check.

The full local 658-test suite and all GitHub checks pass, but the PR adds only good-font fixtures, so none of these rejection paths are covered.

behdad:
I did a review with codex and addressed the findings with Claude. Should be in better shape now, but more LLM review might be justified.

PR #330, opened 2026-09-18
[CFF] Limit total charstring ops that a single glyph can execute

jfkthame:
[edit: the first change mentioned here is incorrect and has been dropped]

This makes two changes to address a performance cliff (potential DoS) when a CFF charstring calls subroutines many times (and at several levels of recursion) to result in executing many millions of charstring ops.

First, we track subroutines that have already been executed successfully, and don't re-execute them if used multiple times. This greatly improves performance on such a font.

Second, we impose a (somewhat arbitrary) limit on the total number of ops that a single glyph's charstring is allowed to execute, and reject the table if this is exceeded. This is desirable because even after the first commit here, which addresses the exponentially-slow performance in OTS, such a font will also have unacceptable performance in any downstream rasterizer that attempts to use it. I don't think there is any legitimate use case for a charstring that expands to many millions of ops.

A couple of test fonts are added: one that fails the total op limit, and a similar but less extreme one that passes.

(Without the patches here, the fonts both pass, but the "bad" one takes many seconds to process.)


jfkthame:
Hmm, this doesn't actually work reliably, because of subrs making changes to the state of the stack. Reconsidering...

jfkthame:
OK, the first change described here has been dropped as being incorrect, so now we just impose a charstring ops execution limit but don't attempt to skip already-seen subroutine execution.

